PR #2, opened 2026-04-07
feat(tracking): add application pipeline tracker with DB persistence

StevSant:
## Summary

- Add new `tracking` module — first ORM model and Alembic migration in the project
- `TrackedApplication` model with status pipeline: saved → applied → interviewing → offered → accepted/rejected
- Full REST API: POST, GET, GET/:id, PATCH, DELETE at `/api/tracking`
- Applications can be linked to ingested jobs (by job_id) or entered manually
- Auto-sets `applied_at` timestamp when status moves to "applied"
- Sync SQLAlchemy repository with PostgreSQL (psycopg2) for persistence
- Added `get_job_by_id` to IngestionOrchestrator for job lookup from tracking
- New "Pipeline" page in frontend with status filter, inline status editing, add form, and delete
- "Track" button on ingestion page for quick one-click job tracking
- 44 new backend unit tests (177 total passing), frontend builds clean

## Test plan

- [x] All 177 backend unit tests pass (`uv run pytest tests/ -v`)
- [x] Frontend builds clean (`npm run build`)
- [ ] Manual test: create tracked application via POST, verify in GET
- [ ] Manual test: update status to "applied", verify applied_at is set
- [ ] Manual test: track job from ingestion page, verify appears in Pipeline
- [ ] Manual test: run Alembic migration against PostgreSQL (`alembic upgrade head`)


StevSant (commented):
## Code Quality Audit Report

### Confirmed Correct

**Architecture and wiring** — The hexagonal dependency rule is respected. The tracking module's domain layer imports only from `kernel/` and its own models; infrastructure depends inward; `main.py` wires everything. No cross-module imports from other bounded contexts.

**Port stub pattern** — `tracking/api/dependencies.py` raises `NotImplementedError` as the DI stub, matching the pattern used in every other module. Consistent.

**Alembic migration** — `001_create_tracked_applications.py` correctly creates the table with `gen_random_uuid()` server default, adds both indexes (`status`, `job_id`), and provides a clean `downgrade()`. The migration matches the ORM model definition.

**`registry.py` pattern** — Importing the ORM model in `infrastructure/registry.py` to ensure `Base.metadata` is populated before Alembic runs is the correct solution for auto-detection.

**`applied_at` idempotency** — The `update_status` service guard (`if status == APPLIED and app.applied_at is None`) correctly prevents overwriting the original application timestamp on re-entry to that status. The test that validates round-tripping through INTERVIEWING and back to APPLIED covers this explicitly.

**PATCH no-op handling** — When neither `status` nor `notes` is supplied, the route falls back to a `service.get(id)` rather than crashing or returning stale data. Defensively correct.

**Test coverage** — The unit test suite is comprehensive: service, repository (SQLite in-memory), routes (fake service), schemas, and models all have dedicated test files with happy-path and edge-case coverage.

**Frontend signal pattern** — The component uses Angular signals throughout (`signal()`, `.update()`, `.set()`) consistently with the rest of the app.

---

### Issues Found

#### CRITICAL

**[BUG] `tracking/domain/models.py` lines 39-47 — `Mapped[None]` is wrong for datetime columns**

`applied_at`, `created_at`, and `updated_at` are all typed as `Mapped[None]`, which tells SQLAlchemy's typed ORM that these columns always return `None`. At runtime, `app.created_at` and `app.updated_at` will be `None` regardless of what the database contains. `TrackedApplicationResponse.model_validate(app)` will serialize those fields as `null`, violating the response schema (both are non-optional `datetime` in the Pydantic schema).

Fix:
```python
from datetime import datetime

applied_at: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), nullable=True)
created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now())
updated_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), onupdate=func.now())
```

Note: the unit tests set `created_at`/`updated_at` manually in fake repositories, which is why this bug is invisible to the current test suite.

---

**[SECURITY] `tracking/api/routes.py` — All five endpoints are unauthenticated**

No route in the tracking module uses `require_auth` from `identity/api/dependencies.py`. The auth guard exists in the project but is not applied to any feature module — tracking continues that pattern. Since the Angular auth guard and HTTP interceptor protect the UI, practical risk is lower, but the API is fully open to anyone who can reach port 8000. Write endpoints (POST/PATCH/DELETE) should at minimum depend on `require_auth`:

```python
from hiresense.identity.api.dependencies import require_auth

@router.post("", ..., dependencies=[Depends(require_auth)])
```

---

#### WARNING

**[BUG] `main.py` line 149 — URL replacement for sync engine is fragile**

`settings.database_url.replace("+asyncpg", "")` hardcodes the assumption that `DATABASE_URL` always uses the asyncpg dialect specifier. If someone configures `postgresql://...` (no driver suffix), the replacement is a no-op and psycopg2 still receives it (which may work). If a different async driver is used (e.g., `+aiopg`), the replacement silently produces a broken URL. A safer approach is a separate `SYNC_DATABASE_URL` env var or explicit URL construction.

---

**[BUG] `tracking/domain/services.py` lines 27, 44 — `.value` passed where enum member is expected**

`track_job` and `track_from_ingestion` both pass `ApplicationStatus.SAVED.value` (the raw string `"saved"`) to the `status` constructor argument. The ORM column is mapped as `Mapped[ApplicationStatus]` so the Python type expects an enum member. With `String(20)` as the column type, SQLAlchemy stores and returns the raw string without coercion — meaning `app.status == ApplicationStatus.SAVED` will evaluate to `False` after a round-trip through the database. Fix: pass `status=ApplicationStatus.SAVED` directly, or switch the column type to `Enum(ApplicationStatus)` for automatic coercion.

---

**[BUG] `tracking/api/routes.py` lines 72-82 — PATCH handler uses a catch-all `ValueError` mapped to 404**

Any `ValueError` raised inside the PATCH block is mapped to 404, but not all `ValueError`s from the service layer are not-found errors. The `create_application` route already demonstrates the better pattern: inspect the exception message for `"not found"` vs `"already tracked"` before deciding the status code. Apply the same defensiveness to the PATCH handler, or introduce a typed exception hierarchy.

---

**[QUALITY] `tracked-application.model.ts` — Two exports in one file violates one-type-per-file rule**

Both `ApplicationStatus` (union type) and `TrackedApplication` (interface) are defined in `tracked-application.model.ts`. Per the project convention, `ApplicationStatus` should live in its own `application-status.type.ts` file.

---

#### INFO

**[QUALITY] Constructor params typed as `Any` in service and repository**

`TrackingService.__init__` and `TrackingRepository.__init__` both use `Any`. Define a `TrackingRepositoryPort` Protocol in `kernel/ports/` or use the concrete type directly. For the orchestrator, `IngestionOrchestrator` can be imported without violating the hexagonal boundary since tracking already depends on it at wiring time.

**[PERFORMANCE] No pagination on `GET /tracking`**

`list_all()` issues an unbounded `SELECT *` with no `LIMIT`. Fine for a personal tracker, but worth noting for when the table grows.

**[A11Y] Delete button has no contextual accessible label**

`<button class="btn-danger">Delete</button>` gives screen reader users no context about which application is being deleted. Prefer `aria-label="Delete {{ app.title }} at {{ app.company }}"`.

**[UX/BUG] `tracking.component.ts` line 80 — Validation condition allows one empty field through**

`if (!title && !company)` only blocks submission when both fields are empty. The backend returns a 422 when either is missing, but the frontend shows a generic error. Use `if (!title || !company)` for proper client-side validation.

**[i18n] Hardcoded English strings throughout the tracking UI**

"Pipeline", "All Statuses", "New Application", "Loading applications...", etc. are hardcoded English. Per project convention, all UI text should go through i18next. This is consistent with the existing pages rather than a regression introduced by this PR, but the pattern should not continue.

---

### Verified

- Migration schema aligns with ORM model (columns, types, nullability, indexes all match)
- Alembic `env.py` correctly imports `registry` before autogenerate runs
- `ingestion_orchestrator._jobs` is populated during `run()`, so `track_from_ingestion` lookup works within a single process lifetime (ephemeral by design — restarts clear the in-memory job store)
- Repository `save()` uses `session.merge()` correctly for re-attaching a detached instance to a new session
- Tracking routes are defined as sync `def`, so FastAPI correctly offloads them to a thread pool — the blocking sync SQLAlchemy calls do not block the event loop
- Frontend `trackedJobIds` signal correctly handles 409 conflicts by marking the job as already tracked

---

### Verdict: PASS WITH FIXES REQUIRED

The feature is well-structured, architecturally consistent, and ships meaningful test coverage. Two issues must be resolved before merge:

1. Fix the `Mapped[None]` type annotations on `applied_at`, `created_at`, `updated_at` — this is a silent data bug where those fields always serialize as `null`.
2. Apply `require_auth` to the write endpoints (POST/PATCH/DELETE) — the tracking API currently has no access control.

The enum-vs-string inconsistency (passing `.value` instead of the enum member) is also worth fixing before merge to avoid subtle comparison bugs.

---

*Review by Claude Code*

PR #3, opened 2026-04-07
feat(matching): add 10-dimension LLM-powered evaluation system

StevSant:
## Summary

- Add 6 new LLM-powered evaluation dimensions: seniority fit, compensation, growth potential, culture fit, application strength, interview readiness
- Each dimension is an independent scorer class extending `BaseLLMScorer` with structured JSON parsing, markdown extraction, regex fallback, and graceful error handling
- `DimensionScorer` protocol + `DimensionResult` model with score clamping (0-1) and weighted composites
- New `POST /api/matching/evaluate` endpoint runs all scorers in parallel via `asyncio.gather` and returns a weighted composite score with per-dimension breakdown
- Wire Anthropic LLM adapter in app factory (graceful fallback to `llm=None` if no API key)
- CV-dependent scorers (application strength, interview readiness) return 0.5 default when no profile provided
- 10 configurable weights via env vars (must sum to 100), defaults updated from 4-dimension to 10-dimension system
- Frontend "Run 10-Dimension Evaluation" button on matching page with dimension cards showing score bars, rationale, and weights
- 48 new backend tests (225 total passing), frontend builds clean

## Test plan

- [x] All 225 backend unit tests pass (`uv run pytest tests/ -v`)
- [x] Frontend builds clean (`npm run build`)
- [ ] Manual test: evaluate a job with LLM API key set, verify 6 new dimensions return real scores
- [ ] Manual test: evaluate without API key, verify all dimensions return 0.5 with "LLM not configured"
- [ ] Manual test: evaluate without profile, verify CV-dependent dimensions return 0.5


StevSant (commented):
## Code Quality Audit — PR #3: 10-Dimension LLM Evaluation System

### Confirmed Correct

The following areas were verified and are working as intended:

- **Weight arithmetic**: All 10 default weights (15+20+10+5+10+10+5+5+10+10) correctly sum to 100. `config.py` and `.env.example` are in sync.
- **Score clamping**: `DimensionResult.clamp_score` correctly guards the [0.0, 1.0] range via `@field_validator`. Tests for over/under-range values pass correctly.
- **LLM response parsing chain**: The three-stage fallback in `BaseLLMScorer._parse_response` (JSON → markdown block → regex float → 0.5) is sensible and well-tested. The markdown regex uses `re.DOTALL` correctly.
- **asyncio.gather usage**: Running all scorers concurrently in `MatchingOrchestrator.evaluate` is the right pattern. The `safe_score` wrapper prevents one failing scorer from cancelling the others.
- **DivisionByZero guard**: `total_weight > 0` fallback in the composite calculation is present and correct.
- **LLM adapter wiring**: The conditional import of `anthropic` at startup gracefully handles missing installs without crashing.
- **One-file-per-class rule**: All six scorer files, model files, and schema files each contain a single class — fully compliant with the project convention.
- **Test coverage**: All new scorers have dedicated unit tests covering happy path, no-LLM fallback, malformed response, and missing profile cases. `test_evaluate_route.py` covers the HTTP layer with a `FakeOrchestrator`.
- **No hardcoded secrets**: API key comes from `settings.llm_api_key` as required.

---

### Issues Found

#### Critical

**[BUG] main.py:164 — Private attribute mutation is fragile and the evaluate route depends on it via getattr**

```python
matching_orchestrator._dimension_scorers = dimension_scorers  # monkey-patch
```
```python
# routes.py:40
dimension_scorers=getattr(orchestrator, "_dimension_scorers", []),
```

`_dimension_scorers` is set via direct attribute assignment on an instance that has no such attribute declared in `__init__`. If a future refactor passes a fresh `MatchingOrchestrator` through the DI system (or a dependency override in a test), `getattr` silently returns `[]`, every scorer is skipped, and the endpoint returns a 0.5 composite with zero dimensions — a silent data-quality failure with no error raised.

**Fix**: Accept `dimension_scorers` in `MatchingOrchestrator.__init__` as a proper parameter and remove the `getattr` call from the route handler.

---

**[BUG] routes.py:39 — profile=None is always passed; two scorers silently default to 0.5**

`ApplicationStrengthScorer` and `InterviewReadinessScorer` both short-circuit to `DimensionResult.default(rationale="No CV provided")` when `profile is None`. Since the `/evaluate` endpoint always passes `profile=None`, these two scorers — weighted at 10 each — permanently return 0.5 regardless of the LLM. The UI displays this as a real score with no indication it is a stub.

This means the composite score is systematically anchored toward 0.5 and the endpoint that advertises 10-dimension evaluation only runs 8 live dimensions.

**Fix**: Either remove those two scorers from `main.py` until profile data is available in the request, or extend `EvaluateRequest` with optional CV fields and construct a profile object from them.

---

#### Warning

**[QUALITY] schemas.py:12 — Bare list default is inconsistent with the codebase convention**

`skills: list[str] = []` is technically safe in Pydantic v2 but `MatchResult` in `models.py` uses `Field(default_factory=list)` for all list fields. Use `Field(default_factory=list)` for consistency.

---

**[QUALITY] matching.component.ts:74 — Job title extracted from first line of description textarea**

```typescript
job_title: this.jobDescription().split('\n')[0] || 'Unknown',
```

The first line of a pasted job description is almost never the job title. The "title" sent to the LLM scorers will be the opening sentence of the description body, producing misleading seniority/compensation scores. A dedicated `jobTitle` input would be more reliable.

---

**[QUALITY] seniority_scorer.py — Developer persona hard-coded into system prompt**

The `SeniorityScorer` system prompt hard-codes "mid-senior backend/AI engineer with 3-5 years of experience" as the evaluation target. This is the developer's personal profile, not a configurable parameter. Every user of the application receives seniority scores calibrated to this specific persona. The persona should come from the `profile` argument or a scorer-level config parameter.

---

**[BUG] matching.component.html:151 — [class] receives hex color strings, not CSS class names**

```typescript
scoreColor(score: number): string {
  if (score >= 0.7) return '#16a34a';
  ...
}
```
```html
[class]="scoreColor(dim.score)"
```

`[class]="expr"` replaces the entire class list with the expression value. The return values are hex color codes, not class names. This sets `class="#16a34a"` on the element — an invalid class name with no visual effect. The dimension bar fill color will not render at all.

**Fix**: Either return CSS class names (e.g., `'bg-green-600'`) and keep `[class]`, or switch to `[style.background-color]="scoreColor(dim.score)"` with the current hex strings.

---

**[i18n] matching.component.html — All new UI strings are hardcoded English**

"Evaluating...", "Run 10-Dimension Evaluation", "Composite Score", "Weight: X%" — none pass through the project's i18n layer. Per project convention, all UI text must go through translation keys.

---

#### Info

**[QUALITY] llm_scorer.py:578 — Regex float fallback can match bare integers in narrative text**

The pattern `r"\b(0\.\d+|1\.0|0|1)\b"` matches the digit `0` or `1` anywhere in a response (e.g., "Step 1: evaluate the role"). A match on a bare `1` in an explanation sentence sets the score to 1.0. Consider removing the `|0|1` alternatives or narrowing the pattern.

---

**[QUALITY] llm_scorer.py:543 — Raw exception message is returned to API clients**

```python
rationale=f"Evaluation failed: {exc}",
```

The raw exception string (potentially including LLM API error payloads) is stored in `DimensionResult.rationale` and returned to the client in the API response. Log at `WARNING` (already done) but return a generic user-facing rationale like `"Evaluation temporarily unavailable"`.

---

**[A11Y] matching.component.html:130 — Evaluate button missing aria-busy during loading**

The button shows "Evaluating..." when loading but has no `aria-busy="true"` attribute. Screen readers rely on this attribute to announce async state changes.

---

### Verified

- Weights sum to 100 confirmed (15+20+10+5+10+10+5+5+10+10 = 100).
- All six scorer files, three frontend model files, and the schemas file are one-class-per-file, compliant with project convention.
- `_parse_response` markdown regex handles both triple-backtick-json and bare triple-backtick blocks via `(?:json)?`.
- `asyncio.gather` relies on the `safe_score` wrapper for per-scorer isolation rather than `return_exceptions=True` — this is correct.
- No hardcoded secrets, API keys, or configurable thresholds in application code.
- `.env.example` updated with all 6 new weight keys and placeholder values.

---

### Verdict: PASS WITH WARNINGS

The scorer architecture is clean, test coverage is thorough, and the LLM plumbing is well-designed. Two issues need attention before this merges:

1. **The private attribute monkey-patch** (`_dimension_scorers`) in `main.py` should be replaced with a proper constructor parameter — it silently returns empty results if the DI wiring ever shifts.
2. **The always-None profile** means `ApplicationStrengthScorer` and `InterviewReadinessScorer` are permanently returning stub scores. Either wire real profile data or remove those scorers until they can be fully activated.

The `[class]` / hex-color bug is a guaranteed UI rendering failure for the dimension bars — fix it alongside the i18n pass.

Review by Claude Code

PR #4, opened 2026-04-07
feat(interview): add story bank and LLM-powered interview prep

StevSant:
## Summary

- New `interview` module with STAR+Reflection story bank persisted to PostgreSQL (second ORM model + Alembic migration)
- Full CRUD REST API for stories (`POST/GET/PATCH/DELETE /api/interview/stories`) with competency filtering
- LLM-powered interview preparation (`POST /api/interview/prepare`) that matches stories to jobs, suggests competencies to probe, technical topics to review, and negotiation talking points
- Graceful degradation when LLM unavailable or story bank empty
- Auth required on all interview endpoints
- New "Interview" page in frontend with story bank management (add/list/delete) and interview prep generation
- 35 new backend tests (260 total passing), frontend builds clean

## Test plan

- [x] All 260 backend unit tests pass
- [x] Frontend builds clean
- [ ] Manual test: add STAR stories, verify persistence
- [ ] Manual test: generate interview prep for a job, verify story matching
- [ ] Manual test: verify prep works with empty story bank

### DIFF
--- a/backend/alembic/versions/002_create_stories.py
+++ b/backend/alembic/versions/002_create_stories.py
@@ -1,0 +1,37 @@
+"""create stories table
+
+Revision ID: 002
+Revises: 001
+Create Date: 2026-04-07
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "002"
+down_revision: Union[str, None] = "001"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "stories",
+        sa.Column("id", sa.Uuid(), primary_key=True, server_default=sa.text("gen_random_uuid()")),
+        sa.Column("title", sa.String(255), nullable=False),
+        sa.Column("competency", sa.String(30), nullable=False),
+        sa.Column("situation", sa.Text(), nullable=False),
+        sa.Column("task", sa.Text(), nullable=False),
+        sa.Column("action", sa.Text(), nullable=False),
+        sa.Column("result", sa.Text(), nullable=False),
+        sa.Column("reflection", sa.Text(), nullable=True),
+        sa.Column("tags", sa.String(500), nullable=True),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=True),
+        sa.Column("updated_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_table("stories")
--- a/backend/src/hiresense/infrastructure/registry.py
+++ b/backend/src/hiresense/infrastructure/registry.py
@@ -1,3 +1,4 @@
 """Import all ORM models so Base.metadata is fully populated for Alembic."""
 
+from hiresense.interview.domain.models import Story  # noqa: F401
 from hiresense.tracking.domain.models import TrackedApplication  # noqa: F401
--- a/backend/src/hiresense/interview/api/dependencies.py
+++ b/backend/src/hiresense/interview/api/dependencies.py
@@ -1,0 +1,9 @@
+from hiresense.interview.domain.services import InterviewPrepService, StoryService
+
+
+def get_story_service() -> StoryService:
+    raise NotImplementedError("Must be overridden during app startup via dependency_overrides")
+
+
+def get_interview_prep_service() -> InterviewPrepService:
+    raise NotImplementedError("Must be overridden during app startup via dependency_overrides")
--- a/backend/src/hiresense/interview/api/routes.py
+++ b/backend/src/hiresense/interview/api/routes.py
@@ -75,7 +75,7 @@ def update_story(
     request: UpdateStoryRequest,
     service: StoryService = Depends(get_story_service),
 ) -> StoryResponse:
-    fields = {k: v for k, v in request.model_dump().items() if v is not None}
+    fields = request.model_dump(exclude_unset=True)
     try:
         story = service.update(id, **fields)
     except ValueError as exc:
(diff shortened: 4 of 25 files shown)


StevSant (commented):
## Code Quality Audit — PR #4: Interview Prep Feature

---

### Confirmed Correct

- Auth enforced at router level via `Depends(require_auth)` — no route is exposed unauthenticated.
- `InterviewPrepService.prepare()` handles three failure modes correctly: no LLM configured, JSON parse failure (including markdown-fenced JSON), and general LLM exception.
- Repository uses scoped sessions via `session_factory()` context manager — no leaked connections.
- `StoryService.update()` delegates the entity-fetch to `self.get()`, so the 404 path is consistently covered.
- Test coverage is thorough: unit tests for models, service, repository (in-memory SQLite), routes (FastAPI TestClient), and prep service. All critical branches are exercised.
- Lazy-loaded route in Angular router is correct — `loadComponent` pattern used consistently.
- Auth interceptor already attaches Bearer token to all outgoing HTTP requests.
- `onupdate=func.now()` on the ORM model is correct for SQLAlchemy Python-side change tracking.

---

### Issues Found

**[CRITICAL — Security] No per-user data isolation on stories**

`backend/src/hiresense/interview/infrastructure/repository.py` — `list_all`, `get_by_id`, `delete`, and `save` have no `user_id` filter. `require_auth` returns the `sub` claim, but neither the route handlers nor the service use it. Any authenticated user can read, update, or delete any other user's stories. The `Story` model also has no `user_id` column. This is broken access control (OWASP A01).

Fix: add `user_id: Mapped[str]` to `Story`, add the column to the Alembic migration, filter every repository query by `user_id`, and thread the `sub` value from `require_auth` into service calls via `Depends`.

---

**[HIGH — Bug] `save()` does not `refresh()` after `merge()` — stale `updated_at` returned to caller**

`backend/src/hiresense/interview/infrastructure/repository.py` lines 33–36

After `session.commit()`, the server-side `updated_at` has been set by the DB, but the in-memory object is never refreshed. The returned Story serializes the old timestamp. `create()` already does this correctly with `session.refresh(story)`.

Fix: add `session.refresh(story)` after `session.commit()` in `save()`.

---

**[HIGH — Bug] PATCH silently drops intentional null/empty clears**

`backend/src/hiresense/interview/api/routes.py` line 78

```python
fields = {k: v for k, v in request.model_dump().items() if v is not None}
```

If a client sends `{"reflection": null}` to clear an optional field, this filter drops it and the update is a no-op. Users can never clear `reflection` or `tags` once set.

Fix: use `request.model_dump(exclude_unset=True)` instead, and allow `None` values to pass through.

---

**[HIGH — Bug] `updated_at` auto-update trigger missing from Alembic migration**

`backend/alembic/versions/002_create_stories.py` line 32

The migration only sets `server_default`. The ORM `onupdate=func.now()` is a SQLAlchemy Python-side hook — it does not create a DB-level trigger. Direct SQL updates (admin tooling, future raw queries) will not update the timestamp.

Fix: add a `BEFORE UPDATE` trigger via `op.execute()` in the migration for PostgreSQL, or document the limitation explicitly.

---

**[MEDIUM — Security] LLM-returned `story_id` values are not validated against the actual story bank**

`backend/src/hiresense/interview/domain/services.py` line 124

`matched_stories` is built directly from raw LLM output. The LLM can hallucinate arbitrary UUIDs. After per-user isolation is added, this becomes a cross-user data leak vector.

Fix: filter the parsed `matched_stories` list to only include IDs present in the `stories` list fetched for this user's session.

---

**[MEDIUM — Security] Raw exception message leaked in API response body**

`backend/src/hiresense/interview/domain/services.py` line 138

```python
negotiation_points=[f"Preparation failed: {exc}"]
```

Raw exception text is serialized into the response. This is an information disclosure risk (internal paths, API error details, etc.).

Fix: log the exception server-side and return a generic message such as `"Preparation unavailable. Please try again."`.

---

**[MEDIUM — Quality] Repository and service types use `Any` — no interface contract**

`backend/src/hiresense/interview/domain/services.py` lines 59 and 67

Both `StoryService.__init__(repository: Any)` and `InterviewPrepService.__init__(llm, story_repo)` lack type annotations. This defeats static analysis and makes the dependency contract invisible.

Fix: define a `Protocol` for the repository interface and add type hints for `llm`.

---

**[LOW — Quality] One-class-per-file rule violated**

`backend/src/hiresense/interview/domain/services.py` — contains `StoryMatch`, `InterviewPrep`, `StoryService`, and `InterviewPrepService`.
`backend/src/hiresense/interview/domain/models.py` — contains `Competency` and `Story`.

Per project convention, each class should be in its own file.

---

**[LOW — i18n] All UI strings are hardcoded English**

`frontend/src/app/pages/interview/interview.component.html` — labels, button text, empty states, and section headers are hardcoded strings. Per project convention, Spanish is the default language and all UI text must go through i18next.

---

**[LOW — UX] No delete confirmation guard on story removal**

`frontend/src/app/pages/interview/interview.component.html` — single click fires an irreversible DELETE. A STAR story can be wiped by misclick.

Fix: add a `confirm()` guard or two-step confirmation UI before issuing the delete request.

---

**[LOW — Performance] Story list fetched without pagination and without user scope**

`backend/src/hiresense/interview/domain/services.py` line 93 — `self._story_repo.list_all()` loads the full bank into memory to build the LLM prompt. No limit is applied.

---

### Verified

- `require_auth` is applied at router scope, not opt-in per route.
- `create()` correctly calls `session.refresh(story)` — the stale-timestamp bug is isolated to `save()`.
- SQLite in-memory fixture in `test_repository.py` tears down cleanly via `Base.metadata.drop_all`.
- Alembic `down_revision = "001"` chain is correct.
- Frontend `deleteStory` filters the local signal list on success, avoiding a full reload.
- Angular `authInterceptor` covers the new interview routes without additional changes.

---

### Verdict: NEEDS FIXES

The critical broken access control (shared story pool across all users) and the two data-correctness bugs (`save()` returning stale `updated_at`, PATCH silently ignoring null clears) must be resolved before merge. LLM story_id validation and exception leakage are high-priority follow-ons.

*Review by Claude Code*

PR #5, opened 2026-04-07
feat(matching): add batch evaluation with ranked leaderboard

StevSant:
## Summary

- New `BatchEvaluationService` with `asyncio.Semaphore` for job-level concurrency control (configurable via `BATCH_CONCURRENCY` env var, default 3)
- `POST /api/matching/batch-evaluate` endpoint that resolves jobs from tracked applications (DB) and/or in-memory ingested jobs, runs 10-dimension evaluation in parallel, returns ranked leaderboard sorted by composite score
- Graceful per-job error handling (failed evaluations get 0.5 score, don't break the batch)
- `IngestionOrchestrator.list_jobs()` method for batch access to ingested jobs
- "Evaluate All" button on Pipeline page with expandable leaderboard showing rank, composite score, source badge, and per-dimension breakdown bars
- 10 new backend tests (270 total passing), frontend builds clean

## Test plan

- [x] All 270 backend unit tests pass
- [x] Frontend builds clean
- [ ] Manual test: batch evaluate tracked applications, verify sorted leaderboard
- [ ] Manual test: batch evaluate with `include_ingested: true`, verify both sources appear
- [ ] Manual test: verify concurrency cap works (check LLM call timing)


StevSant (commented):
## Code Quality Audit — PR #5: Batch Evaluation + Leaderboard

---

### Confirmed Correct

**Semaphore instantiation in `__init__`** — `asyncio.Semaphore(concurrency)` created at startup and reused across calls is correct. A common mistake is creating it inside `evaluate_batch()` per call; this avoids that.

**`asyncio.gather` + per-task exception isolation** — Each `evaluate_one` coroutine catches its own exception and returns a fallback `BatchResult` instead of propagating. `gather` therefore always resolves fully, so callers never receive a partial result silently mixed with an exception object. Solid pattern.

**Sorted leaderboard descending** — `sorted(..., key=lambda r: r.composite_score, reverse=True)` is correct and done after gather completes, not inside individual coroutines. No race on the sort.

**Config follows project conventions** — `batch_concurrency: int = 3` in `Settings`, documented in `.env.example` under its own section. No hardcoded value in application code.

**DI wiring order in `main.py`** — Cross-module overrides (`get_tracking_service_for_matching`, `get_ingestion_orchestrator_for_matching`) are registered after both the tracking and ingestion singletons are fully constructed. No ordering hazard.

**Frontend models follow one-interface-per-file convention** — `batch-evaluation-response.model.ts` and `batch-result.model.ts` are each a single interface. Correct.

**Tests cover the key paths** — Empty list, specific IDs, all-tracked, with-ingested, missing ID skip, failure fallback, concurrency cap, and sort order are all tested. The concurrency test with `SlowOrchestrator` using an async lock to track `max_concurrent` is a clean approach.

---

### Issues Found

**[CRITICAL — BUG] `batch_service.py:51` — Failure fallback `composite_score=0.5` poisons the leaderboard**

When any individual job evaluation throws, the fallback `BatchResult` is returned with `composite_score=0.5`. This value participates in the sort and will appear ranked in the leaderboard alongside successful results, silently misrepresenting a failed evaluation as a mid-range match. A user has no way to distinguish a genuine 50% score from a failure stub.

Fix: add a `failed: bool = False` field to `BatchResult` and `BatchResultResponse`, set it `True` in the except branch, and either filter these from the sort or render them visually distinct in the frontend. Alternatively, use `composite_score=0.0` so failures always sink to the bottom and are obviously wrong.

---

**[CRITICAL — BUG] `routes.py:96` — Tracked apps sent to the LLM with `description: ""`**

Both the all-tracked and specific-IDs branches hard-code `"description": ""` when building the job dict. `TrackedApplication` carries no description field so an empty string is the only option — but the LLM scorers receive a job with almost nothing to work with beyond title and company name. Dimensions like `culture_fit`, `growth_potential`, and `seniority_fit` will produce structurally unreliable scores for tracked apps while ingested jobs (which carry full descriptions) score more accurately. The leaderboard mix of the two sources will mislead users.

Fix (minimum): use `"description": app.notes or ""` to give the LLM the user-entered freeform context. Better: surface this in the response with a `data_quality` field or a frontend disclaimer for `source: "tracked"` results.

---

**[WARNING — BUG] `batch_service.py:51` — `0.5` is a hardcoded configurable threshold**

Separate from the leaderboard-poisoning concern: the sentinel value `0.5` in application logic violates the project no-hardcoded-values rule. If scoring weights change, 0.5 may no longer represent a meaningful midpoint.

Fix: use `0.0` (unambiguous failure floor), or extract it as a module-level named constant.

---

**[WARNING — QUALITY] `batch_service.py` — Two classes in one file violates one-class-per-file rule**

`BatchResult` (a Pydantic model) and `BatchEvaluationService` both live in `batch_service.py`. Project convention is one class per file.

Fix: move `BatchResult` to `matching/domain/models.py` alongside `MatchResult`, or create a dedicated `batch_result.py`. The service stays in `batch_service.py`.

---

**[WARNING — QUALITY] `routes.py:87–123` — Route handler contains orchestration logic that belongs in the service**

The `batch_evaluate` route manually assembles the `jobs` list by iterating the tracking service and applying conditional logic (specific IDs vs all, include_ingested). This is orchestration that belongs in `BatchEvaluationService` or a dedicated assembler, not an API handler. The job-assembly path in the route is not directly unit-tested — `FakeBatchService.evaluate_batch` in `test_batch_route.py` echoes jobs back without exercising the assembly conditions.

Fix: extract job-dict assembly into a method on `BatchEvaluationService` (e.g., `build_job_list(tracked_app_ids, tracking_service, ingestion_orchestrator, include_ingested)`) and test it in `test_batch_service.py`.

---

**[WARNING — A11Y / i18n] `tracking.component.html` and `tracking.component.ts` — All new UI text is hardcoded English**

Strings added by this PR are hardcoded English: "Leaderboard", "Evaluating X jobs...", "Evaluate All", "Seniority Fit", "Compensation", etc. Project rule: all UI text goes through i18next with Spanish as the default language. This is a recurring violation.

Affected: `tracking.component.html` lines 88, 142; `tracking.component.ts` lines 175–183.

Fix: add strings to the translation file under a `tracking` or `pipeline` namespace and use the translate pipe or `TranslateService`.

---

**[WARNING — A11Y] `tracking.component.html:146` — Interactive `div` is not keyboard accessible**

The leaderboard card uses `<div class="leaderboard-card" (click)="toggleExpand(result.source_id)">` with no `tabindex`, no `role`, and no `keydown` handler. Keyboard users cannot expand dimensions.

Fix:
```html
<div class="leaderboard-card"
     role="button"
     tabindex="0"
     (click)="toggleExpand(result.source_id)"
     (keydown.enter)="toggleExpand(result.source_id)"
     (keydown.space)="toggleExpand(result.source_id)"
     [attr.aria-expanded]="expandedResultId() === result.source_id">
```

---

**[INFO] `routes.py:80` — `batch-evaluate` is auth-gated while `evaluate` and `analyze` remain public**

The new endpoint correctly applies `require_auth` because it reads user pipeline data. The existing `/matching/evaluate` and `/matching/analyze` are intentionally public (they operate on caller-supplied data, not stored user data). The distinction is correct and intentional — worth a comment on the unprotected endpoints to prevent future confusion.

---

**[INFO] `batch_service.py:35` — `profile=None` always passed to orchestrator**

`evaluate(job=job, profile=None)` ignores any uploaded CV/profile for batch evaluation, same as the single `/evaluate` endpoint. Not a regression, but batch evaluation is the primary comparative ranking surface, so the gap is more visible here. Worth a `# TODO: thread user profile into batch evaluation` comment.

---

**[INFO — SCSS] `tracking.component.scss:707` — Leaderboard cards use `border-radius: 8px` instead of project-standard 12px**

Project card convention is `rounded-xl` / 12px. New `.leaderboard-card` uses 8px. Minor visual inconsistency.

---

### Verified

- `list_jobs()` returns `list(self._jobs.values())` — a snapshot copy, not a live dict reference. Safe.
- `BatchEvaluateRequest.tracked_app_ids` is `list[uuid_mod.UUID]` — Pydantic rejects malformed UUIDs before the route executes.
- `tracking_service.get(app_id)` raises `ValueError` on missing ID; route catches `ValueError` and continues. Correct exception type match.
- `angular.json` bump from 4kB to 6kB style warning threshold is expected given the new SCSS block.
- `FakeTrackingService.list()` accepts `status=None` matching the real `TrackingService.list(status=None)` signature.

---

### Verdict: PASS WITH WARNINGS

The core mechanics (semaphore, gather, sort, DI wiring, config) are solid and well-tested. Two issues need attention before merge: the failure fallback `0.5` silently pollutes leaderboard rankings, and tracked apps evaluated with empty descriptions will produce unreliable scores that users may trust. The one-class-per-file violation and missing i18n are recurring project conventions that should be addressed. The keyboard accessibility gap on the leaderboard card is a quick fix.

Review by Claude Code

PR #6, opened 2026-04-08
feat(research): add LLM-powered company deep research

StevSant:
## Summary

- New `research` bounded context with CompanyResearch ORM model, Alembic migration (003), sync repository with case-insensitive lookup
- `CompanyResearchService` with DB caching: `research()` (cache-first), `refresh()` (force re-research), `get()` (cache-only)
- LLM prompt includes company name + optional job description for richer context
- Graceful degradation: no-LLM and LLM-failure return unpersisted fallbacks so next request retries
- Three REST endpoints: `POST /research`, `POST /research/refresh`, `GET /research/{company_name}` (all auth-gated)
- Inline "Research" button on Pipeline page with collapsible detail panel showing funding, tech stack, culture, growth, pros/cons, red flags
- "Refresh" button + last-updated timestamp per company
- 19 new backend tests (289 total passing), frontend builds clean

## Test plan

- [x] All 289 backend unit tests pass
- [x] Frontend builds clean
- [ ] Manual test: research a company, verify LLM response parsed and displayed
- [ ] Manual test: research same company again, verify cache hit (no LLM call)
- [ ] Manual test: refresh research, verify updated data
- [ ] Manual test: verify graceful fallback without LLM API key

### DIFF
--- a/backend/alembic/versions/003_create_company_research.py
+++ b/backend/alembic/versions/003_create_company_research.py
@@ -1,0 +1,40 @@
+"""create company_research table
+
+Revision ID: 003
+Revises: 002
+Create Date: 2026-04-07
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "003"
+down_revision: Union[str, None] = "002"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "company_research",
+        sa.Column("id", sa.Uuid(), primary_key=True, server_default=sa.text("gen_random_uuid()")),
+        sa.Column("company_name", sa.String(255), nullable=False),
+        sa.Column("funding_stage", sa.String(100), nullable=False),
+        sa.Column("tech_stack", sa.Text(), nullable=False),
+        sa.Column("culture_summary", sa.Text(), nullable=False),
+        sa.Column("growth_trajectory", sa.Text(), nullable=False),
+        sa.Column("red_flags", sa.Text(), nullable=True),
+        sa.Column("pros", sa.Text(), nullable=False),
+        sa.Column("cons", sa.Text(), nullable=False),
+        sa.Column("raw_llm_response", sa.Text(), nullable=False),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=True),
+        sa.Column("updated_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=True),
+    )
+    op.create_index("ix_company_research_company_name", "company_research", ["company_name"], unique=True)
+
+
+def downgrade() -> None:
+    op.drop_index("ix_company_research_company_name")
+    op.drop_table("company_research")
--- a/backend/alembic/versions/003_create_company_research.py
+++ b/backend/alembic/versions/003_create_company_research.py
@@ -36,5 +36,5 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    op.drop_index("ix_company_research_company_name")
+    op.drop_index("ix_company_research_company_name", table_name="company_research")
     op.drop_table("company_research")
--- a/backend/src/hiresense/infrastructure/registry.py
+++ b/backend/src/hiresense/infrastructure/registry.py
@@ -1,4 +1,5 @@
 """Import all ORM models so Base.metadata is fully populated for Alembic."""
 
 from hiresense.interview.domain.models import Story  # noqa: F401
+from hiresense.research.domain.models import CompanyResearch  # noqa: F401
 from hiresense.tracking.domain.models import TrackedApplication  # noqa: F401
--- a/backend/src/hiresense/main.py
+++ b/backend/src/hiresense/main.py
@@ -59,6 +59,10 @@ from hiresense.interview.api.dependencies import get_story_service, get_intervie
 from hiresense.interview.api.routes import router as interview_router
 from hiresense.interview.domain.services import InterviewPrepService, StoryService
 from hiresense.interview.infrastructure.repository import StoryRepository
+from hiresense.research.infrastructure.repository import CompanyResearchRepository
+from hiresense.research.domain.services import CompanyResearchService
+from hiresense.research.api.dependencies import get_company_research_service
+from hiresense.research.api.routes import router as research_router
 
 
 def create_app() -> FastAPI:
@@ -209,6 +213,12 @@ def create_app() -> FastAPI:
     app.dependency_overrides[get_story_service] = lambda: story_service
     app.dependency_overrides[get_interview_prep_service] = lambda: interview_prep_service
     app.include_router(interview_router)
+
+    # --- Research module ---
+    research_repo = CompanyResearchRepository(session_factory=sync_session_factory)
+    research_service = CompanyResearchService(llm=llm, repository=research_repo)
+    app.dependency_overrides[get_company_research_service] = lambda: research_service
+    app.include_router(research_router)
 
     # --- Health check ---
     @app.get("/health")
--- a/backend/src/hiresense/research/api/dependencies.py
+++ b/backend/src/hiresense/research/api/dependencies.py
@@ -1,0 +1,5 @@
+from __future__ import annotations
+
+
+def get_company_research_service():
+    raise NotImplementedError("Must be overridden during app startup via dependency_overrides")
--- a/backend/src/hiresense/research/api/routes.py
+++ b/backend/src/hiresense/research/api/routes.py
@@ -9,20 +9,20 @@ router = APIRouter(prefix="/research", tags=["research"], dependencies=[Depends(
 
 
 @router.post("", response_model=CompanyResearchResponse)
-def research_company(
+async def research_company(
     request: ResearchRequest,
     service: CompanyResearchService = Depends(get_company_research_service),
 ) -> CompanyResearchResponse:
-    result = service.research(company_name=request.company_name, job_description=request.job_description)
+    result = await service.research(company_name=request.company_name, job_description=request.job_description)
     return CompanyResearchResponse.model_validate(result)
 
 
 @router.post("/refresh", response_model=CompanyResearchResponse)
-def refresh_research(
+async def refresh_research(
     request: ResearchRequest,
     service: CompanyResearchService = Depends(get_company_research_service),
 ) -> CompanyResearchResponse:
-    result = service.refresh(company_name=request.company_name, job_description=request.job_description)
+    result = await service.refresh(company_name=request.company_name, job_description=request.job_description)
     return CompanyResearchResponse.model_validate(result)
 
 
--- a/backend/src/hiresense/research/api/routes.py
+++ b/backend/src/hiresense/research/api/routes.py
@@ -1,0 +1,37 @@
+from __future__ import annotations
+from fastapi import APIRouter, Depends, HTTPException
+from hiresense.identity.api.dependencies import require_auth
+from hiresense.research.api.dependencies import get_company_research_service
+from hiresense.research.api.schemas import CompanyResearchResponse, ResearchRequest
+from hiresense.research.domain.services import CompanyResearchService
+
+router = APIRouter(prefix="/research", tags=["research"], dependencies=[Depends(require_auth)])
+
+
+@router.post("", response_model=CompanyResearchResponse)
+def research_company(
+    request: ResearchRequest,
+    service: CompanyResearchService = Depends(get_company_research_service),
+) -> CompanyResearchResponse:
+    result = service.research(company_name=request.company_name, job_description=request.job_description)
+    return CompanyResearchResponse.model_validate(result)
+
+
+@router.post("/refresh", response_model=CompanyResearchResponse)
+def refresh_research(
+    request: ResearchRequest,
+    service: CompanyResearchService = Depends(get_company_research_service),
+) -> CompanyResearchResponse:
+    result = service.refresh(company_name=request.company_name, job_description=request.job_description)
+    return CompanyResearchResponse.model_validate(result)
+
+
+@router.get("/{company_name}", response_model=CompanyResearchResponse)
+def get_research(
+    company_name: str,
+    service: CompanyResearchService = Depends(get_company_research_service),
+) -> CompanyResearchResponse:
+    result = service.get(company_name)
+    if result is None:
+        raise HTTPException(status_code=404, detail=f"No research found for {company_name}")
+    return CompanyResearchResponse.model_validate(result)
--- a/backend/src/hiresense/research/api/schemas.py
+++ b/backend/src/hiresense/research/api/schemas.py
@@ -1,16 +1,16 @@
 from __future__ import annotations
 import uuid
 from datetime import datetime
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 
 class ResearchRequest(BaseModel):
-    company_name: str
+    company_name: str = Field(min_length=1)
     job_description: str = ""
 
 
 class CompanyResearchResponse(BaseModel):
-    id: uuid.UUID
+    id: uuid.UUID | None = None
     company_name: str
     funding_stage: str
     tech_stack: str
--- a/backend/src/hiresense/research/api/schemas.py
+++ b/backend/src/hiresense/research/api/schemas.py
@@ -1,0 +1,25 @@
+from __future__ import annotations
+import uuid
+from datetime import datetime
+from pydantic import BaseModel
+
+
+class ResearchRequest(BaseModel):
+    company_name: str
+    job_description: str = ""
+
+
+class CompanyResearchResponse(BaseModel):
+    id: uuid.UUID
+    company_name: str
+    funding_stage: str
+    tech_stack: str
+    culture_summary: str
+    growth_trajectory: str
+    red_flags: str | None
+    pros: str
+    cons: str
+    created_at: datetime | None = None
+    updated_at: datetime | None = None
+
+    model_config = {"from_attributes": True}
--- a/backend/src/hiresense/research/domain/models.py
+++ b/backend/src/hiresense/research/domain/models.py
@@ -1,0 +1,33 @@
+from __future__ import annotations
+
+import uuid as uuid_mod
+from datetime import datetime
+
+from sqlalchemy import DateTime, Index, String, Text, Uuid, func
+from sqlalchemy.orm import Mapped, mapped_column
+
+from hiresense.infrastructure.database import Base
+
+
+class CompanyResearch(Base):
+    __tablename__ = "company_research"
+    __table_args__ = (
+        Index("ix_company_research_company_name", "company_name", unique=True),
+    )
+
+    id: Mapped[uuid_mod.UUID] = mapped_column(Uuid, primary_key=True, default=uuid_mod.uuid4)
+    company_name: Mapped[str] = mapped_column(String(255), nullable=False)
+    funding_stage: Mapped[str] = mapped_column(String(100), nullable=False)
+    tech_stack: Mapped[str] = mapped_column(Text, nullable=False)
+    culture_summary: Mapped[str] = mapped_column(Text, nullable=False)
+    growth_trajectory: Mapped[str] = mapped_column(Text, nullable=False)
+    red_flags: Mapped[str | None] = mapped_column(Text, nullable=True)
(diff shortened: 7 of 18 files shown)


StevSant (commented):
## Code Quality Audit — PR #6: Company Research

---

### Critical (must fix)

**[BUG] `backend/src/hiresense/research/domain/services.py:36` — LLM call is never awaited; the service is entirely broken at runtime**

`AnthropicLLMAdapter.complete()` is declared `async def complete(...)`. The research service calls it as `response = self._llm.complete(prompt)` with no `await`, and the route handlers are not `async def`. This means `response` receives a coroutine object, not a string. `_parse_response` will immediately raise `ValueError`, the `except Exception` swallows it, and every call silently returns the `"Research unavailable"` fallback. The feature appears to work (HTTP 200) but is always returning placeholder data.

Compare with the interview service, which correctly uses `async def prepare(...)` and `await self._llm.complete(...)`. The fix requires making `_do_research`, `research`, `refresh`, `get` all `async def`, and making the three route handlers `async def` as well.

```python
# services.py — correct pattern
async def _do_research(self, company_name: str, job_description: str) -> CompanyResearch:
    ...
    response = await self._llm.complete(prompt)

# routes.py — correct pattern
@router.post("", response_model=CompanyResearchResponse)
async def research_company(...) -> CompanyResearchResponse:
    result = await service.research(...)
```

---

**[BUG] `backend/alembic/versions/003_create_company_research.py:39` — `downgrade()` will fail on PostgreSQL**

`op.drop_index("ix_company_research_company_name")` is missing the required `table_name` argument. Migrations 001 and 002 both pass `table_name=` and that is the pattern required by Alembic when targeting databases that use schema-qualified index names (PostgreSQL). On PostgreSQL this raises `CompileError` at downgrade time, making the migration partially irreversible.

Fix: `op.drop_index("ix_company_research_company_name", table_name="company_research")`

---

### Warning (should fix)

**[BUG] `frontend/src/app/pages/tracking/tracking.component.html:139` / `tracking.component.ts:190` — "View Research" button re-fetches instead of toggling**

When cached research already exists, the button label changes to "View Research", but the click handler still calls `researchCompany(app)` which unconditionally fires `POST /research`. The `toggleResearch(appId)` method exists in the component but is never wired up in the template. As a result, every "View Research" click makes a redundant network round-trip even though the data is already cached locally.

Fix: change the `@else if` branch to call `toggleResearch(app.id)` and stop event propagation there instead of letting the outer `researchCompany` handler fire.

---

**[QUALITY] `backend/src/hiresense/research/domain/services.py:61` — bare `except Exception` silently swallows all errors including programming mistakes**

The broad catch conceals `KeyError` (if the LLM returns JSON missing a required field), `TypeError`, and even the coroutine-not-awaited bug described above. This makes the service very hard to debug in production — failures are invisible unless you check the fallback sentinel string in the response body.

Fix: catch specific expected exceptions (`json.JSONDecodeError`, `ValueError`, `KeyError`) separately, and either re-raise or log unexpected exceptions before returning the fallback.

---

**[RELIABILITY] `backend/src/hiresense/research/domain/services.py:36` — no timeout on LLM call**

`self._llm.complete(prompt)` has no per-call timeout. A slow or hung Anthropic API call will block the request indefinitely. The interview service has the same gap, but this endpoint is especially exposed because company research queries tend to generate longer prompts.

Fix: wrap the call with `asyncio.wait_for(self._llm.complete(prompt), timeout=settings.llm_timeout)` and handle `asyncio.TimeoutError` explicitly.

---

**[SECURITY] `backend/src/hiresense/research/domain/services.py:67` — company name injected directly into LLM prompt without sanitization**

`company_name` comes directly from user input via `ResearchRequest` and is interpolated into the prompt with no sanitization. A user could supply a string with embedded newlines or instruction-injection text. While blast radius is limited here (no tool use, no sensitive data in scope), the pattern should be hardened.

Fix: add `company_name: str = Field(min_length=1, max_length=200)` in `ResearchRequest`, and strip or reject strings containing newlines before prompt interpolation.

---

**[RELIABILITY] `backend/src/hiresense/research/api/schemas.py:8` — no input length constraints on `company_name`**

`company_name: str` with no minimum length accepts an empty string, which produces a nonsensical LLM query and a DB record with `company_name = ""`.

Fix:
```python
from pydantic import Field

class ResearchRequest(BaseModel):
    company_name: str = Field(min_length=1, max_length=200)
    job_description: str = Field(default="", max_length=10000)
```

---

**[RELIABILITY] `backend/src/hiresense/research/domain/services.py:97-108` — fallback object returned without explicit `id` may be `None` at serialization time**

`_make_fallback` constructs a `CompanyResearch(...)` without setting `id`. Although `id` has `default=uuid_mod.uuid4`, SQLAlchemy only fires Python-side column defaults when the object is flushed into a session. An in-memory-only instance may have `id=None`. The response schema declares `id: uuid.UUID` as non-optional, so `model_validate` would raise a `ValidationError`, turning graceful degradation into a 500.

Fix: explicitly assign `id=uuid_mod.uuid4()` in `_make_fallback`.

---

**[PERFORMANCE] `backend/src/hiresense/research/domain/services.py:38` — second DB lookup inside `_do_research` is always redundant during `refresh`**

After the LLM call succeeds, `_do_research` calls `self._repo.get_by_company_name(company_name)` again. When invoked via `refresh()`, this is a wasted round-trip because no prior lookup was performed. The result is two DB reads per refresh when one would suffice.

Fix: accept an optional `existing: CompanyResearch | None` parameter in `_do_research` and pass it through from the callers.

---

### Info (consider fixing)

**[A11Y] `frontend/src/app/pages/tracking/tracking.component.html:134` — Research button has no contextual `aria-label`**

The button text "Research" / "View Research" gives no indication of which company row it acts on. Screen readers will announce "Research, button" with no company context.

Fix: add `[attr.aria-label]="'Research ' + app.company"` to the button element.

---

**[QUALITY] `backend/src/hiresense/research/infrastructure/repository.py:29` — `save()` via `session.merge()` issues an implicit SELECT before the UPDATE**

`merge` on a detached instance must re-populate the identity map, adding an extra DB round-trip on every refresh. Not a correctness issue at current scale, but worth noting if this endpoint sees frequent refreshes.

---

### Confirmed Correct

- Auth guard applied at the router level via `dependencies=[Depends(require_auth)]` — all three endpoints are protected.
- Case-insensitive lookup via `func.lower()` on both sides is correct and portable across backends.
- The unique index on `company_name` is consistent between the ORM model and the Alembic migration.
- Prompt length-capping `job_description[:2000]` prevents runaway token usage.
- `raw_llm_response` is correctly excluded from the API response schema — raw LLM output is not leaked to clients.
- Test coverage is solid: model, repository (SQLite in-memory), service, and route layers all have dedicated test files with meaningful cases covering cache-hit, LLM failure, case-insensitive lookup, and not-found scenarios.

---

### Summary

2 critical, 6 warnings, 2 info

**Verdict: NEEDS FIXES** — the async/await mismatch means the feature does not work end-to-end with the real LLM at all (every call returns the fallback silently). The missing `table_name` in `downgrade()` will cause a PostgreSQL migration failure. Both must be resolved before merge.

---

Review by Claude Code

StevSant (commented):
## Code Quality Audit — PR #6: Company Research

---

### Critical (must fix)

**[BUG] `backend/src/hiresense/research/domain/services.py:36` — LLM call is never awaited; the service is entirely broken at runtime**

`AnthropicLLMAdapter.complete()` is declared `async def complete(...)`. The research service calls it as `response = self._llm.complete(prompt)` with no `await`, and the route handlers are not `async def`. This means `response` receives a coroutine object, not a string. `_parse_response` will immediately raise `ValueError`, the `except Exception` swallows it, and every call silently returns the `"Research unavailable"` fallback. The feature appears to work (HTTP 200) but is always returning placeholder data.

Compare with the interview service, which correctly uses `async def prepare(...)` and `await self._llm.complete(...)`. The fix requires making `_do_research`, `research`, `refresh`, `get` all `async def`, and making the three route handlers `async def` as well.

```python
# services.py — correct pattern
async def _do_research(self, company_name: str, job_description: str) -> CompanyResearch:
    ...
    response = await self._llm.complete(prompt)

# routes.py — correct pattern
@router.post("", response_model=CompanyResearchResponse)
async def research_company(...) -> CompanyResearchResponse:
    result = await service.research(...)
```

---

**[BUG] `backend/alembic/versions/003_create_company_research.py:39` — `downgrade()` will fail on PostgreSQL**

`op.drop_index("ix_company_research_company_name")` is missing the required `table_name` argument. Migrations 001 and 002 both pass `table_name=` and that is the pattern required by Alembic when targeting databases that use schema-qualified index names (PostgreSQL). On PostgreSQL this raises `CompileError` at downgrade time, making the migration partially irreversible.

Fix: `op.drop_index("ix_company_research_company_name", table_name="company_research")`

---

### Warning (should fix)

**[BUG] `frontend/src/app/pages/tracking/tracking.component.html:139` / `tracking.component.ts:190` — "View Research" button re-fetches instead of toggling**

When cached research already exists, the button label changes to "View Research", but the click handler still calls `researchCompany(app)` which unconditionally fires `POST /research`. The `toggleResearch(appId)` method exists in the component but is never wired up in the template. As a result, every "View Research" click makes a redundant network round-trip even though the data is already cached locally.

Fix: change the `@else if` branch to call `toggleResearch(app.id)` and stop event propagation there instead of letting the outer `researchCompany` handler fire.

---

**[QUALITY] `backend/src/hiresense/research/domain/services.py:61` — bare `except Exception` silently swallows all errors including programming mistakes**

The broad catch conceals `KeyError` (if the LLM returns JSON missing a required field), `TypeError`, and even the coroutine-not-awaited bug described above. This makes the service very hard to debug in production — failures are invisible unless you check the fallback sentinel string in the response body.

Fix: catch specific expected exceptions (`json.JSONDecodeError`, `ValueError`, `KeyError`) separately, and either re-raise or log unexpected exceptions before returning the fallback.

---

**[RELIABILITY] `backend/src/hiresense/research/domain/services.py:36` — no timeout on LLM call**

`self._llm.complete(prompt)` has no per-call timeout. A slow or hung Anthropic API call will block the request indefinitely. The interview service has the same gap, but this endpoint is especially exposed because company research queries tend to generate longer prompts.

Fix: wrap the call with `asyncio.wait_for(self._llm.complete(prompt), timeout=settings.llm_timeout)` and handle `asyncio.TimeoutError` explicitly.

---

**[SECURITY] `backend/src/hiresense/research/domain/services.py:67` — company name injected directly into LLM prompt without sanitization**

`company_name` comes directly from user input via `ResearchRequest` and is interpolated into the prompt with no sanitization. A user could supply a string with embedded newlines or instruction-injection text. While blast radius is limited here (no tool use, no sensitive data in scope), the pattern should be hardened.

Fix: add `company_name: str = Field(min_length=1, max_length=200)` in `ResearchRequest`, and strip or reject strings containing newlines before prompt interpolation.

---

**[RELIABILITY] `backend/src/hiresense/research/api/schemas.py:8` — no input length constraints on `company_name`**

`company_name: str` with no minimum length accepts an empty string, which produces a nonsensical LLM query and a DB record with `company_name = ""`.

Fix:
```python
from pydantic import Field

class ResearchRequest(BaseModel):
    company_name: str = Field(min_length=1, max_length=200)
    job_description: str = Field(default="", max_length=10000)
```

---

**[RELIABILITY] `backend/src/hiresense/research/domain/services.py:97-108` — fallback object returned without explicit `id` may be `None` at serialization time**

`_make_fallback` constructs a `CompanyResearch(...)` without setting `id`. Although `id` has `default=uuid_mod.uuid4`, SQLAlchemy only fires Python-side column defaults when the object is flushed into a session. An in-memory-only instance may have `id=None`. The response schema declares `id: uuid.UUID` as non-optional, so `model_validate` would raise a `ValidationError`, turning graceful degradation into a 500.

Fix: explicitly assign `id=uuid_mod.uuid4()` in `_make_fallback`.

---

**[PERFORMANCE] `backend/src/hiresense/research/domain/services.py:38` — second DB lookup inside `_do_research` is always redundant during `refresh`**

After the LLM call succeeds, `_do_research` calls `self._repo.get_by_company_name(company_name)` again. When invoked via `refresh()`, this is a wasted round-trip because no prior lookup was performed. The result is two DB reads per refresh when one would suffice.

Fix: accept an optional `existing: CompanyResearch | None` parameter in `_do_research` and pass it through from the callers.

---

### Info (consider fixing)

**[A11Y] `frontend/src/app/pages/tracking/tracking.component.html:134` — Research button has no contextual `aria-label`**

The button text "Research" / "View Research" gives no indication of which company row it acts on. Screen readers will announce "Research, button" with no company context.

Fix: add `[attr.aria-label]="'Research ' + app.company"` to the button element.

---

**[QUALITY] `backend/src/hiresense/research/infrastructure/repository.py:29` — `save()` via `session.merge()` issues an implicit SELECT before the UPDATE**

`merge` on a detached instance must re-populate the identity map, adding an extra DB round-trip on every refresh. Not a correctness issue at current scale, but worth noting if this endpoint sees frequent refreshes.

---

### Confirmed Correct

- Auth guard applied at the router level via `dependencies=[Depends(require_auth)]` — all three endpoints are protected.
- Case-insensitive lookup via `func.lower()` on both sides is correct and portable across backends.
- The unique index on `company_name` is consistent between the ORM model and the Alembic migration.
- Prompt length-capping `job_description[:2000]` prevents runaway token usage.
- `raw_llm_response` is correctly excluded from the API response schema — raw LLM output is not leaked to clients.
- Test coverage is solid: model, repository (SQLite in-memory), service, and route layers all have dedicated test files with meaningful cases covering cache-hit, LLM failure, case-insensitive lookup, and not-found scenarios.

---

### Summary

2 critical, 6 warnings, 2 info

**Verdict: NEEDS FIXES** — the async/await mismatch means the feature does not work end-to-end with the real LLM at all (every call returns the fallback silently). The missing `table_name` in `downgrade()` will cause a PostgreSQL migration failure. Both must be resolved before merge.

---

Review by Claude Code

StevSant (commented):
## Code Quality Audit — PR #6: Company Research

---

### Critical (must fix)

**[BUG] `backend/src/hiresense/research/domain/services.py:36` — LLM call is never awaited; the service is entirely broken at runtime**

`AnthropicLLMAdapter.complete()` is declared `async def complete(...)`. The research service calls it as `response = self._llm.complete(prompt)` with no `await`, and the route handlers are not `async def`. This means `response` receives a coroutine object, not a string. `_parse_response` will immediately raise `ValueError`, the `except Exception` swallows it, and every call silently returns the `"Research unavailable"` fallback. The feature appears to work (HTTP 200) but is always returning placeholder data.

Compare with the interview service, which correctly uses `async def prepare(...)` and `await self._llm.complete(...)`. The fix requires making `_do_research`, `research`, `refresh`, `get` all `async def`, and making the three route handlers `async def` as well.

```python
# services.py — correct pattern
async def _do_research(self, company_name: str, job_description: str) -> CompanyResearch:
    ...
    response = await self._llm.complete(prompt)

# routes.py — correct pattern
@router.post("", response_model=CompanyResearchResponse)
async def research_company(...) -> CompanyResearchResponse:
    result = await service.research(...)
```

---

**[BUG] `backend/alembic/versions/003_create_company_research.py:39` — `downgrade()` will fail on PostgreSQL**

`op.drop_index("ix_company_research_company_name")` is missing the required `table_name` argument. Migrations 001 and 002 both pass `table_name=` and that is the pattern required by Alembic when targeting databases that use schema-qualified index names (PostgreSQL). On PostgreSQL this raises `CompileError` at downgrade time, making the migration partially irreversible.

Fix: `op.drop_index("ix_company_research_company_name", table_name="company_research")`

---

### Warning (should fix)

**[BUG] `frontend/src/app/pages/tracking/tracking.component.html:139` / `tracking.component.ts:190` — "View Research" button re-fetches instead of toggling**

When cached research already exists, the button label changes to "View Research", but the click handler still calls `researchCompany(app)` which unconditionally fires `POST /research`. The `toggleResearch(appId)` method exists in the component but is never wired up in the template. As a result, every "View Research" click makes a redundant network round-trip even though the data is already cached locally.

Fix: change the `@else if` branch to call `toggleResearch(app.id)` and stop event propagation there instead of letting the outer `researchCompany` handler fire.

---

**[QUALITY] `backend/src/hiresense/research/domain/services.py:61` — bare `except Exception` silently swallows all errors including programming mistakes**

The broad catch conceals `KeyError` (if the LLM returns JSON missing a required field), `TypeError`, and even the coroutine-not-awaited bug described above. This makes the service very hard to debug in production — failures are invisible unless you check the fallback sentinel string in the response body.

Fix: catch specific expected exceptions (`json.JSONDecodeError`, `ValueError`, `KeyError`) separately, and either re-raise or log unexpected exceptions before returning the fallback.

---

**[RELIABILITY] `backend/src/hiresense/research/domain/services.py:36` — no timeout on LLM call**

`self._llm.complete(prompt)` has no per-call timeout. A slow or hung Anthropic API call will block the request indefinitely. The interview service has the same gap, but this endpoint is especially exposed because company research queries tend to generate longer prompts.

Fix: wrap the call with `asyncio.wait_for(self._llm.complete(prompt), timeout=settings.llm_timeout)` and handle `asyncio.TimeoutError` explicitly.

---

**[SECURITY] `backend/src/hiresense/research/domain/services.py:67` — company name injected directly into LLM prompt without sanitization**

`company_name` comes directly from user input via `ResearchRequest` and is interpolated into the prompt with no sanitization. A user could supply a string with embedded newlines or instruction-injection text. While blast radius is limited here (no tool use, no sensitive data in scope), the pattern should be hardened.

Fix: add `company_name: str = Field(min_length=1, max_length=200)` in `ResearchRequest`, and strip or reject strings containing newlines before prompt interpolation.

---

**[RELIABILITY] `backend/src/hiresense/research/api/schemas.py:8` — no input length constraints on `company_name`**

`company_name: str` with no minimum length accepts an empty string, which produces a nonsensical LLM query and a DB record with `company_name = ""`.

Fix:
```python
from pydantic import Field

class ResearchRequest(BaseModel):
    company_name: str = Field(min_length=1, max_length=200)
    job_description: str = Field(default="", max_length=10000)
```

---

**[RELIABILITY] `backend/src/hiresense/research/domain/services.py:97-108` — fallback object returned without explicit `id` may be `None` at serialization time**

`_make_fallback` constructs a `CompanyResearch(...)` without setting `id`. Although `id` has `default=uuid_mod.uuid4`, SQLAlchemy only fires Python-side column defaults when the object is flushed into a session. An in-memory-only instance may have `id=None`. The response schema declares `id: uuid.UUID` as non-optional, so `model_validate` would raise a `ValidationError`, turning graceful degradation into a 500.

Fix: explicitly assign `id=uuid_mod.uuid4()` in `_make_fallback`.

---

**[PERFORMANCE] `backend/src/hiresense/research/domain/services.py:38` — second DB lookup inside `_do_research` is always redundant during `refresh`**

After the LLM call succeeds, `_do_research` calls `self._repo.get_by_company_name(company_name)` again. When invoked via `refresh()`, this is a wasted round-trip because no prior lookup was performed. The result is two DB reads per refresh when one would suffice.

Fix: accept an optional `existing: CompanyResearch | None` parameter in `_do_research` and pass it through from the callers.

---

### Info (consider fixing)

**[A11Y] `frontend/src/app/pages/tracking/tracking.component.html:134` — Research button has no contextual `aria-label`**

The button text "Research" / "View Research" gives no indication of which company row it acts on. Screen readers will announce "Research, button" with no company context.

Fix: add `[attr.aria-label]="'Research ' + app.company"` to the button element.

---

**[QUALITY] `backend/src/hiresense/research/infrastructure/repository.py:29` — `save()` via `session.merge()` issues an implicit SELECT before the UPDATE**

`merge` on a detached instance must re-populate the identity map, adding an extra DB round-trip on every refresh. Not a correctness issue at current scale, but worth noting if this endpoint sees frequent refreshes.

---

### Confirmed Correct

- Auth guard applied at the router level via `dependencies=[Depends(require_auth)]` — all three endpoints are protected.
- Case-insensitive lookup via `func.lower()` on both sides is correct and portable across backends.
- The unique index on `company_name` is consistent between the ORM model and the Alembic migration.
- Prompt length-capping `job_description[:2000]` prevents runaway token usage.
- `raw_llm_response` is correctly excluded from the API response schema — raw LLM output is not leaked to clients.
- Test coverage is solid: model, repository (SQLite in-memory), service, and route layers all have dedicated test files with meaningful cases covering cache-hit, LLM failure, case-insensitive lookup, and not-found scenarios.

---

### Summary

2 critical, 6 warnings, 2 info

**Verdict: NEEDS FIXES** — the async/await mismatch means the feature does not work end-to-end with the real LLM at all (every call returns the fallback silently). The missing `table_name` in `downgrade()` will cause a PostgreSQL migration failure. Both must be resolved before merge.

---

Review by Claude Code

PR #9, opened 2026-05-25
Application pipeline spec + LinkedIn data + tracking PATCH fix

StevSant:
## Problema

- LinkedIn job detail panel renders mostly empty values (description, posted date, seniority, etc.) because the adapter only parsed search-card fields and never fetched the per-job detail page.
- `PATCH /tracking/{id}` returns 500 with `DetachedInstanceError` on `updated_at` — the SQLAlchemy session closes before Pydantic reads the server-side default. Blocks status changes on tracked applications.
- The product has five vertical slices (ingestion → matching → optimization → tracking → interview) but no spine connecting them. Optimization asks the user to manually type required/missing skills that the matching step already computed. `TrackedApplication` is a stub with no link to match/optimized CV/interview prep. Created applications don't surface on the Interview page.

## Cambio

- **LinkedIn adapter (`backend/src/hiresense/ingestion/adapters/linkedin.py`)**: parses `<time datetime>` from the card for `posted_date`; fetches each job's detail page from LinkedIn's guest `jobPosting` endpoint behind a `Semaphore(3)` with per-request delay; extracts description from `div.show-more-less-html__markup` plus the criteria block (seniority level / employment type / job function / industries). Detail-fetch failures are logged per-job and don't abort the batch.
- **LinkedIn normalizer**: surfaces description, parses `posted_date` to `datetime`, maps `job_function` → `department`, packs seniority + employment type into `categories`.
- **`TrackingRepository.save()`**: adds `session.refresh(application)` after commit, mirroring the existing `create()` path so `updated_at` is loaded before the session closes.
- **Spec doc (`docs/superpowers/specs/2026-05-24-application-pipeline-redesign-design.md`)**: Approach B design — `TrackedApplication` becomes the spine; four child artifact tables (job snapshot 1:1; matches/CV optimizations/interview preps 1:N). New `hiresense/applications/` bounded context with full REST surface. Skill auto-extraction (ingested → copy `NormalizedJob.skills`; manual → LLM extract). Missing-skills computed live from snapshot vs profile, persisted on each match row. Frontend gets a unified `/dashboard/applications/:id` detail view with Job / Match / CV / Interview tabs; Tracking renamed to Applications; Interview page surfaces tracked applications at the top.

Also included in this PR (already committed earlier on `main` and rolled into this branch): an earlier ingestion design spec and the job-detail-panel implementation plan.

## Resuelve / Impacto

- LinkedIn detail panel now renders real data on next ingestion run (description, posted date, seniority, employment type).
- `PATCH /tracking/{id}` returns 200 cleanly; status changes work end-to-end.
- Application pipeline redesign is unblocked: implementation plan can be drafted against the committed spec.

## Test plan

- [ ] Re-run a LinkedIn ingestion via `POST /ingestion/fetch` and open the detail panel of a LinkedIn row — verify description, posted date, seniority, and employment type are populated.
- [ ] PATCH a tracked application's status from the UI — verify 200 response (previously 500).
- [ ] Read through `docs/superpowers/specs/2026-05-24-application-pipeline-redesign-design.md` end-to-end.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/src/hiresense/ingestion/adapters/linkedin.py
+++ b/backend/src/hiresense/ingestion/adapters/linkedin.py
@@ -1,13 +1,17 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 import re
 from typing import Any
 
 from bs4 import BeautifulSoup
 
+from hiresense.ingestion.domain.html_stripper import strip_html
 from hiresense.ingestion.domain.models import RawJobListing
 from hiresense.kernel.value_objects import SourceType
+
+logger = logging.getLogger(__name__)
 
 USER_AGENT = (
     "Mozilla/5.0 (Windows NT 10.0; Win64; x64) "
@@ -17,6 +21,8 @@ USER_AGENT = (
 MAX_PAGES = 4
 PAGE_SIZE = 25
 REQUEST_DELAY = 2.0
+DETAIL_CONCURRENCY = 3
+DETAIL_DELAY = 0.3
 
 
 class LinkedInAdapter:
@@ -36,7 +42,7 @@ class LinkedInAdapter:
         keywords = filters.get("keywords", "software engineer") if filters else "software engineer"
         location = filters.get("location", "") if filters else ""
         headers = {"User-Agent": USER_AGENT}
-        jobs: list[RawJobListing] = []
+        cards: list[dict[str, str]] = []
         for page in range(MAX_PAGES):
             start = page * PAGE_SIZE
             url = f"{self._base_url}/seeMoreJobPostings/search"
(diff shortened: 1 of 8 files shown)


StevSant (commented):
## Code Review (Claude Code)

### Confirmed Correct
- `TrackingRepository.save()` fix is structurally sound. `session.merge()` returns a session-bound instance; `session.refresh()` on that instance, before the `with` block closes, reloads server defaults like `updated_at`. Matches the existing `create()` path. No double-refresh or detached-entity risk.
- 429 handling on the LinkedIn detail endpoint is present (status_code check before `raise_for_status`), and pagination already handles 429 the same way.
- `_parse_date` correctly silences malformed ISO dates with a bare `except ValueError`, returning `None` as the right sentinel for an optional field.

### Issues Found

- **[important] `linkedin.py:125` — `return_exceptions=True` silently discards enrichment failures.** `fetch_one` already catches every exception inside `_fetch_detail`, so exceptions never bubble to `gather`. The flag was redundant and would have masked any future bug in `fetch_one` or the semaphore code. **Fixed in f786d68** — dropped the flag and added an aggregate `"X/Y jobs enriched"` log line so we have visibility when LinkedIn blocks the whole batch.
- **[important] No aggregate visibility when all detail fetches fail.** **Fixed in f786d68** (same commit).
- **[minor] `DETAIL_CONCURRENCY` / `DETAIL_DELAY` are module constants per the no-hardcoded-values rule.** Existing constants (`MAX_PAGES`, `PAGE_SIZE`, `REQUEST_DELAY`) follow the same module-constant pattern, so this matches local convention. Worth moving to `hiresense.config.Settings` later given they're per-environment tuning knobs, but not blocking.
- **[minor] `linkedin_normalizer.py:17` — `posted_date` parsed as naive `datetime`.** `datetime.strptime(..., "%Y-%m-%d")` produces a naive datetime. If the column is timezone-aware (the SQLAlchemy `DateTime(timezone=True)` columns elsewhere in the schema are), this can cause comparison errors. Follow-up: parse as UTC.

### Verified
- `return_exceptions=True` cannot mask the already-wired per-job logging in `_fetch_detail` — failures there are logged individually before the gather sees them.
- `session.refresh` is on the merged (session-bound) instance, not the original passed-in object. Fix is correct.
- `DETAIL_DELAY` is applied inside the semaphore, so the delay is per-slot, not a global barrier. With `DETAIL_CONCURRENCY=3`, three requests fire in parallel each preceded by 0.3s sleep within their slot — intended behaviour for a polite crawler.

### Verdict
**LGTM** — both `important` items have been addressed in f786d68. Minor items (config-ize tuning knobs; timezone-aware `posted_date`) are good follow-ups but not blocking.

Review by Claude Code

Issue #18, opened 2026-05-30
fix(ingestion): best-matching jobs cannot surface — scoring runs after pagination


lesquel:
## Summary

The job ranking is fundamentally broken: **the best-matching jobs cannot reach page 1**. Scoring (semantic + LLM) is computed **after** pagination, so a strong match that lands on page 2/3 is never promoted. On a cold database the global ordering collapses to keyword/skill overlap only, which favors tag-stuffed postings over genuinely relevant roles.

This is the root cause behind "the best jobs don't surface."

## Pipeline today (`GET /ingestion/jobs`)

1. `repository.list_all()` loads all jobs for the bucket with **no `ORDER BY`** (`ingestion/infrastructure/jobs_repository.py:88`).
2. Skill-overlap score is computed for every job; `match_score = combine_fit_score(skill_score, job.semantic_score)`. On cold state `semantic_score is None`, so `match_score` is **skill overlap only** (`ingestion/api/routes.py:145`).
3. `filter_and_paginate()` **sorts here** by `match_score` then slices the page (`ingestion/domain/job_filter.py:136`). ← the ordering that decides page 1 happens with semantic/LLM signals still missing.
4. Semantic scoring runs **only on the already-sliced page** and re-sorts that slice only (`ingestion/api/routes.py:181`).
5. LLM quick scoring also runs **only on the visible page** and re-sorts that slice only (`ingestion/api/routes.py:227`).

### CRITICAL — Semantic scoring happens after pagination
`routes.py:175` sorts/paginates while `semantic_score` is `None` for unviewed jobs; `routes.py:181` scores only `result.jobs` (the slice). A semantically excellent job with low keyword overlap sorts to the bottom and never appears on page 1.

### CRITICAL — LLM quick scoring happens after pagination
`routes.py:227-238` scores only the visible page (comment: "bounded to the visible page"). The highest-quality signal (seniority/core-skill/discipline gating in `quick_scoring_service.py`) cannot promote a job from page 3 to page 1.

### CRITICAL — pgvector ANN search is write-only
`JobEmbeddingIndexer.index()` only `upsert`s vectors. `PgVectorStore.search()` exists (`adapters/vector_store/pgvector_adapter.py:75`, runs `ORDER BY embedding <=> :q LIMIT :k`) but is **never called** in the ranking path. The list endpoint uses the in-memory `SemanticScoringService`, re-embedding per request, lost on restart. The global ANN pre-rank the whole feature was designed for does not exist.

[...]

PR #21, opened 2026-05-30
feat(ingestion): SemanticPreRanker domain service + DI wiring (PR 2/4, #18)

lesquel:
## PR 2 of 4 — `SemanticPreRanker` domain service + DI wiring

Part of the fix for **#18**. Second slice of the 4-PR stack. **Stacked on PR1 (#20)** — review this diff against the PR1 branch; GitHub will retarget to `main` once #20 merges.

> ⚠️ **No behavior change yet.** This PR adds the global pre-ranking *capability* and wires it into the composition root, but does **not** call it from `routes.py`. That wiring — where the observable ranking actually changes — is **PR3**. After this PR merges, the job list still ranks exactly as before.

### What this PR does (Work Units D + E)

- **D — `SemanticPreRanker`** (`ingestion/domain/semantic_pre_ranker.py`, new): a domain service that makes the **entire corpus** rankable by semantic similarity to the candidate profile *before* pagination. It:
  - Embeds the profile (cached, async-locked against thundering-herd) and calls `VectorStorePort.search(profile_vec, top_k=settings.prerank_top_k_cap, filters={"bucket": tab})`.
  - Sets `semantic_score` on every indexed job from the ANN result, recomputes `match_score` via the weighted blend (`combine_fit_score`, 0.4 skill / 0.6 semantic), and sorts the whole corpus by it — so a **high-semantic / low-keyword** job can rank to the top (the core #18 fix).
  - Appends **unindexed** jobs (not yet embedded / outside `top_k`) to the tail preserving order — **no job is ever dropped** (count invariant holds).
  - **Graceful fallback**: if the vector store is absent, `search()` raises, the profile is empty, or the embedding can't be computed (incl. empty-vector guard), it returns the jobs **unchanged** — the list endpoint can never crash because of pre-ranking.
  - Depends **only on ports** (`VectorStorePort` + embedding port) — clean hexagonal layering, no infrastructure imports in domain.
- **E — DI wiring**: instantiated in `bootstrap/ingestion.py`, exposed on `IngestionProvider`, accessible via `get_pre_ranker()` in `dependencies.py`. `routes.py` production code is **untouched** (confirmed empty diff).

### Why split this way

The behavior-changing `routes.py` integration is isolated in PR3 so the reviewer of *that* PR sees only the integration diff, and this PR can be verified purely as "correct, well-tested, inert plumbing."

### Testing

- Strict TDD. **17 `SemanticPreRanker` tests** + 2 wiring tests: ordering by combined score, `semantic_score` populated from ANN, unindexed→tail order preserved, count invariant, `top_k`/bucket forwarded, all fallback paths (None store, raise, empty profile, empty embedding), cache cold/warm, weight injection changes order.
- `cd backend && uv run pytest tests/unit/ingestion` → **243 passed**.
- `ruff check .` → zero new violations.
- Fresh adversarial review run on this slice; the math was independently verified (skill=0.1/semantic=0.95 → 0.61 beats skill=0.9/semantic=0.20 → 0.48). Two review items fixed in `c7bcb23` (dead variable removed, empty-embedding guard added).

### Stack
1. PR1 (#20) — config + weights + batched persistence
2. **PR2 (this)** — SemanticPreRanker + DI wiring
3. PR3 — routes integration (global pre-rank before pagination, default sort) + backfill ← **behavior changes here**
4. PR4 — pgvector integration tests

Part of #18.

### DIFF
--- a/backend/src/hiresense/bootstrap/ingestion.py
+++ b/backend/src/hiresense/bootstrap/ingestion.py
@@ -169,9 +169,21 @@ def build_ingestion(infra: SharedInfra, tracked: Callable[[str], Any]) -> Ingest
         indexer=portals_indexer,
     )
 
+    from hiresense.ingestion.domain.semantic_pre_ranker import SemanticPreRanker
     from hiresense.ingestion.domain.semantic_scoring_service import SemanticScoringService
 
     semantic_scoring = SemanticScoringService(embedding_port=infra.embedding)
+
+    # SemanticPreRanker wires vector store + embedding for global ANN pre-ranking.
+    # When vector store is None (no pgvector configured), pre_ranker is still
+    # constructed — its own passthrough logic handles the None vector_store case.
+    pre_ranker = SemanticPreRanker(
+        infra.vector_store,
+        infra.embedding,
+        top_k_cap=s.prerank_top_k_cap,
+        skill_weight=s.prerank_weight_skill,
+        semantic_weight=s.prerank_weight_semantic,
+    )
 
     match_cache_repo = JobMatchCacheRepository(session_factory=infra.sync_session_factory)
     quick_scoring = QuickScoringService(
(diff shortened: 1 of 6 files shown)


StevSant:
Superseded by #22 (merged). This PR was auto-closed by GitHub when #20's base branch was deleted on merge. #22 is this exact branch merged onto current `main` (job-lifecycle + #20's foundation), conflicts resolved keeping both the pre-ranker and revalidation wiring; full backend suite green (599). PR3/PR4 of #18 can branch off `main` now.

Issue #156, opened 2026-07-06
Security: authInterceptor attaches the bearer token to every request regardless of origin


StevSant:
**Severity:** Low
**Location:** `frontend/src/app/core/interceptors/auth.interceptor.ts:5-14`
**Confidence:** Medium

## Description
The interceptor unconditionally adds `Authorization: Bearer <token>` to every `HttpClient` request with no origin check. Today all calls are same-origin (`/api`), so it is not currently exploitable, but the first time any service hits a third-party URL the session token would be silently sent to that origin.

## Proposed solution
Attach the header only when the URL is relative or matches the app's API base (`req.url.startsWith('/api')` or `environment.apiUrl`).

## Acceptance criteria
- [ ] Token attached only to same-origin/API requests
- [ ] Spec asserts a third-party URL gets no token

**Estimated effort:** XS

StevSant:
Verified against current `main` (post-#186 httpOnly-cookie migration) — already fixed, closing.

The `auth.interceptor.ts` this report targets no longer exists. The cookie migration replaced the bearer-token interceptor with `credentialsInterceptor`, scoped exactly as the proposed solution asks:

- `frontend/src/app/core/interceptors/credentials.interceptor.ts:9-18` — sets `withCredentials` only when `req.url.startsWith(environment.apiUrl)`; third-party origins get nothing. There is no bearer token anywhere in the app anymore (grep for `Authorization`/`Bearer` returns zero non-spec hits).
- `frontend/src/app/core/interceptors/credentials.interceptor.spec.ts:33-40` — already asserts a third-party URL receives no credentials, matching acceptance criterion 2 verbatim.

Both acceptance criteria are met by existing code; no change needed.

Issue #153, opened 2026-07-06
Security: Add a Content-Security-Policy


StevSant:
**Severity:** Low
**Location:** `frontend/src/index.html`; no server CSP header configured
**Confidence:** High

## Description
There is no CSP anywhere. Combined with the localStorage token, any injected inline script or malicious third-party resource runs with full origin privileges. The page loads external Google Fonts, which a CSP would also constrain.

## Proposed solution
Add a strict CSP (ideally a response header from the server/reverse proxy, or a `<meta http-equiv>` fallback): `default-src 'self'`; `connect-src 'self'`; `font-src https://fonts.gstatic.com`; `style-src 'self' https://fonts.googleapis.com`; `object-src 'none'`; `frame-ancestors 'none'`. Verify the blob-URL PDF preview iframe still works.

## Acceptance criteria
- [ ] CSP served (header or meta)
- [ ] App (incl. PDF preview) works under the policy

**Estimated effort:** M
**References:** OWASP Secure Headers Project

PR #186, opened 2026-07-07
fix(identity): move session auth to an httpOnly cookie (#131)

### LINKED ISSUE #131: Security: Move the session token out of localStorage (XSS token theft)
**Severity:** High
**Location:** `frontend/src/app/core/services/auth.service.ts:34,43,48-53`
**Confidence:** High

## Description
The access token is persisted in `localStorage('access_token')`. `localStorage` is fully readable by any JS on the origin, so a single XSS (compromised dependency, injected script) exfiltrates the bearer token and can replay it against the API for the token's full 24h lifetime. An httpOnly/Secure/SameSite cookie is unreachable from JS and removes this class of theft.

## Proposed solution
Set the session token in an httpOnly Secure SameSite=Strict cookie from the backend on login; switch the client to `withCredentials` and drop the localStorage-based `Authorization` interceptor. If cookies are infeasible, add a strict CSP (separate issue) and shorten token lifetime.

## Acceptance criteria
- [ ] Token no longer stored in localStorage/sessionStorage
- [ ] Auth works via httpOnly cookie (or documented compensating controls)
- [ ] Logout clears the cookie server-side

**Estimated effort:** L
**References:** OWASP "HTML5 Security Cheat Sheet — Local Storage", CWE-522

StevSant:
## ⚠️ Decision required

This finding (#131, High/P1) has two mutually exclusive fixes. This PR implements **Option A**; the user should pick at review.

**Option A — httpOnly `SameSite=Strict` cookie (implemented here).**
The backend sets the JWT in an httpOnly, Secure (mode-aware), `SameSite=Strict` cookie on login; the SPA switches to `withCredentials` and no longer stores or reads any token in JS. This **removes the XSS token-theft class entirely** — an injected script can neither read nor exfiltrate the session, because JS has no access to the cookie.

**Option B — keep the bearer token in JS + compensating controls (NOT implemented).**
Keep `localStorage`/in-memory bearer auth but add a strict Content-Security-Policy (tracked as #153) and shorten token lifetime. This only *mitigates* XSS exfiltration (a CSP bypass or a same-context read still leaks the token); it does not eliminate the class.

**Why A was chosen:** it eliminates the vulnerability rather than narrowing the window. CSRF — the usual tradeoff when moving auth into a cookie — is closed by `SameSite=Strict` (the browser never sends the cookie on cross-site requests), so we don't trade one class of bug for another.

**If you prefer Option B, revert:**
- `backend/`: `identity/api/cookies.py` (delete), the cookie set/clear in `identity/api/routes.py` (login/logout), the cookie fallback in `identity/api/dependencies.py` (restore `HTTPBearer()` with `auto_error=True`), and the `session_cookie_*` / `jwt_expiry_hours` config in `config/groups/core.py` + `config/mode.py` + `.env.example`.
- `frontend/`: restore `localStorage` in `auth.service.ts`, restore the token-attaching `auth.interceptor.ts` (revert `credentials.interceptor.ts`), and revert the guards to synchronous `isAuthenticated()`.
- Then pursue #153 (strict CSP) and a shorter `JWT_EXPIRY_HOURS`.

---

## Summary

Moves the session token out of `localStorage` (where any XSS could steal the 24h bearer) into an httpOnly cookie the browser manages.

**Backend**
- Login sets the JWT in an httpOnly, `SameSite=Strict` cookie; Secure is mode-aware (on for `APP_MODE=production`, off for local http). Token is still returned in the body and the `Authorization: Bearer` header is still accepted (fallback for API tooling/tests).
- New `POST /auth/logout` clears the cookie server-side.
- Auth dependencies (`require_auth`/`get_current_user`/`require_admin`) read the token from the header **or** the cookie. A missing token is a 401 before the auth service is touched, so unauthenticated requests don't depend on identity wiring.
- Cookie name / Secure / SameSite and `JWT_EXPIRY_HOURS` (which also drives cookie max-age) are configurable through the config package + `.env.example` — no hardcoded values.

**Frontend**
- `AuthService` no longer stores a token; session state is derived from `/auth/me` (a cached probe) since JS can't read the httpOnly cookie. `isAuthenticated`/`role`/`isAdmin` signals now reflect the server response.
- Replaced the token-attaching interceptor with a `credentialsInterceptor` scoped to the API base (sets `withCredentials`), so the cookie is never offered to third-party origins.
- Guards resolve session state via `ensureLoaded()` before deciding; `logout()` calls the server. Removed the now-dead client-side JWT-decode utilities (`is-token-expired`, `role-from-token`).
- Corrected the admin CSV-export comment: the same-origin `window.open` now authenticates via the cookie automatically (this path was effectively broken under bearer-in-JS, which couldn't attach a header to a plain navigation).

Closes #131

## Decisions taken

- **Bearer header kept as a fallback.** The auth dependency accepts the cookie *and* `Authorization: Bearer`, so existing API tooling and the integration-test conventions keep working. The cookie is the SPA's primary transport.
- **Secure flag is mode-aware.** `SESSION_COOKIE_SECURE` defaults to unset → resolved by `APP_MODE` in `config/mode.apply_mode` (production→on, local→off). Setting it explicitly (`true`/`false`) overrides. Local http dev would otherwise silently drop a Secure cookie.
- **Reused the existing `/auth/me` endpoint** as the SPA's "am I authenticated" probe instead of adding a new one — it already returns `{username, role}`, exactly what the guards need.
- **CSRF stance:** `SameSite=Strict` is the mitigation; no separate CSRF token is introduced. Called out explicitly above.
- **Missing-token → 401 (was 403 via `HTTPBearer` default).** Switching `HTTPBearer(auto_error=False)` + explicit check yields 401 (correct "not authenticated") and matches the existing `test_requires_auth_without_token` assertions.
- **`get_auth_service` returns `AuthService | None`** so bare test apps (no identity wired, no token) 401 cleanly while still honoring `app.dependency_overrides[get_auth_service]` used by route tests.
- **`JWT_EXPIRY_HOURS` config added** to remove the previously hardcoded 24h token TTL and keep the cookie max-age in lock-step.
- **Scope for #156** (interceptor attaching auth to every origin): not fixed here, but not made worse — credentials are scoped to the API base rather than a blanket `withCredentials`.

## Verification

Backend (`backend/`, `UV_LINK_MODE=copy`, dummy `DATABASE_URL` since Settings requires Postgres):
- `uv run python -m pytest` → **1274 passed, 6 skipped**
- `uv run ruff check .` → **All checks passed**
- New `tests/integration/test_cookie_auth.py`: cookie login sets httpOnly/SameSite=strict; Secure only in production mode; cookie-alone auth; bearer-only auth; logout clears + deauthenticates; missing creds → 401. New unit test asserts token TTL honors `JWT_EXPIRY_HOURS`.

Frontend (`frontend/`, Node 22 via fnm):
- `npm test` → **397 passed (76 files)**
- `npm run build` → bundle generated OK
- `npx ng lint` → **All files pass linting** (CI runs this; caught a `type`→`interface` issue that test/build did not)

Traps checked: `.env.example` updated with the new keys + comments; new settings live in the `core` config group (mode-aware Secure resolved in `config/mode.py`); `identity/api/__init__.py` re-exports the new cookie helpers; domain layer imports no framework code (cookie helpers live in `api/`).


StevSant:
### Self-review (full diff `origin/main...HEAD`)

Reviewed for bugs, security gaps, and missed acceptance criteria.

**Acceptance criteria — met**
- ✅ Token no longer in localStorage/sessionStorage (`auth.service.ts` stores nothing; dead JWT-decode utils removed; grep for `access_token`/`setToken`/`getToken` is clean).
- ✅ Auth works via httpOnly cookie (integration test drives cookie-only login → `/auth/me` → protected route).
- ✅ Logout clears the cookie server-side (`POST /auth/logout` → `delete_cookie`; test confirms deauth).

**Security notes**
- The login **response body still contains the token** (needed for the Bearer fallback / API tooling). The SPA discards it, so persistent XSS theft is gone; only a same-instant XSS during the login round-trip could read it — a far narrower window than 24h localStorage, and an accepted tradeoff. If the reviewer wants to fully close even that, we can gate the body token behind a query flag / separate tooling endpoint (follow-up, not #131).
- CSRF: `SameSite=Strict` chosen over `Lax`. Consequence: a session won't carry across a top-level cross-site navigation into the app (e.g. following an external link straight into an authed deep link re-prompts login). Acceptable for a single-user tool; noted so it's a conscious choice.
- `logout` requires no auth by design (idempotent clear). It only ever expires a cookie — no state leak.

**Correctness**
- Auth deps resolve `get_auth_service` via `Depends` (not a direct call), so `app.dependency_overrides[get_auth_service]` in existing route tests still works — this was a regression I caught and fixed mid-implementation.
- No duplicate `/auth/me` on login: `login` → `refreshSession` sets status `authenticated`, so the subsequent `authGuard.ensureLoaded()` short-circuits to `of(true)` without a second probe.
- `errorInterceptor` exempts `/auth/login|me|logout` so the session probe's 401 is handled once by `AuthService`, not double-handled into a redirect loop.
- All admin routes gate through identity's `require_admin` (admin module re-exports it) — no stray `HTTPBearer` bypassing the cookie path.

**Config hygiene:** new keys in the `core` group + `.env.example`; Secure resolved mode-aware in `config/mode.py`; `identity/api/__init__.py` re-exports the cookie helpers; domain layer untouched (cookie helpers are api-layer).

StevSant:
**Decision resolved: Option A (httpOnly cookie) confirmed.** Branch rebased onto main including #188's identity hardening — `AuthService`/`IdentityProvider` now take both `password_hash` (#188) and `expiry_hours` (this PR), and the SPA interceptor chain keeps main's `timeoutInterceptor` with `credentialsInterceptor` replacing the removed `authInterceptor`. Verified post-rebase: backend pytest + ruff clean; frontend 448 tests, lint, and build clean.

PR #187, opened 2026-07-07
fix(security): P1 backend hardening — LaTeX shell-escape, admin password hash, outreach rate limits

### LINKED ISSUE #128: Security: Compile LaTeX with -no-shell-escape to close file-read/RCE surface
**Severity:** High
**Location:** `backend/src/hiresense/adapters/latex/compiler.py:78-93` (`_run_once`); reached via `applications/domain/apply_service.py` (raw `raw_tex`) and `optimization/domain/services.py` (spliced `optimized_tex`)
**Confidence:** High

## Description
`xelatex` is invoked without `-no-shell-escape`. Two untrusted LaTeX inputs reach the compiler: (a) the user's raw profile CV `raw_tex`, compiled verbatim on `?original=true`, and (b) `optimized_tex`, where LLM-produced replacement strings are spliced into the original by `CVOptimizer._apply_changes` (the "only modify content text" instruction is non-binding, and job descriptions feeding the optimizer are attacker-influenceable). Under TeX Live's default *restricted* shell-escape, `\input{/etc/passwd}` / `\openin` file reads are still permitted, so crafted input can exfiltrate server files into the generated PDF; if the image enables full shell-escape, `\write18` is RCE. `docker-compose.yml` runs `APP_MODE=production`, so this is a real hosted-deployment exposure.

## Proposed solution
Pass `-no-shell-escape` to every `xelatex` invocation in `_run_once`. Optionally harden the temp dir with `openin_any=p` / `TEXMFOUTPUT`. Add a regression test compiling a doc containing `\input{...}` and `\write18` and asserting compilation fails.

## Acceptance criteria
- [ ] `-no-shell-escape` present on the xelatex argv
- [ ] Regression test proves `\write18`/`\input` of an absolute path is rejected
- [ ] Both original-CV and optimized-CV compile paths covered

**Estimated effort:** S
**References:** CWE-78, TeX Live shell-escape docs

### LINKED ISSUE #129: Security: Hash the admin password and compare in constant time
**Severity:** High
**Location:** `backend/src/hiresense/identity/domain/services.py:16-19`; credential from `config/groups/core.py`, `bootstrap/identity.py`
**Confidence:** High

## Description
The admin password is stored in plaintext (`AUTH_PASSWORD`) and verified with `password == self._password`. Two problems: (a) no hashing means any leak of `.env`, config, a backup, process memory, or a log line directly exposes a reusable (and likely reused) credential — no defense-in-depth; (b) `==` on `str` is not constant-time, a timing side channel that leaks length/prefix. This is the authentication path for a PII-heavy app.

## Proposed solution
Store an argon2 or bcrypt hash (e.g. `AUTH_PASSWORD_HASH`) and verify via the library's constant-time `verify()`. If plaintext config must remain for the single-user model, at minimum compare both username and password with `secrets.compare_digest(...)` to remove the timing channel. Document the new env var in `.env.example`.

## Acceptance criteria
- [ ] Password verified against a hash (or at least via `secrets.compare_digest`)
- [ ] No plaintext password retained in the service after startup where avoidable
- [ ] `.env.example` updated
- [ ] Unit test for correct/incorrect credential

**Estimated effort:** M
**References:** OWASP ASVS V2.4, CWE-208 (timing), CWE-256 (plaintext storage)

### LINKED ISSUE #130: Security: Rate-limit and validate recipients on /outreach/send (open email relay)
**Severity:** High
**Location:** `backend/src/hiresense/outreach/api/routes.py:23,57-75`; `outreach/api/schemas.py:28-34`; `outreach/domain/outreach_service.py:100-128`
**Confidence:** High

## Description
The outreach router is gated only by `require_auth` (not `require_admin`) and, unlike other LLM/network-heavy endpoints, applies no `enforce_expensive_rate_limit`. `SendRequest.to`/`subject`/`message` are free-form (`to` is a bare `str`, not `EmailStr`) and are not tied to the application's real contact — the only check is that `application_id` exists. Any bearer-token holder (or a stolen/XSS-leaked token) can drive the configured SMTP account to send arbitrary content to arbitrary recipients at unbounded volume: a spam/phishing relay plus uncapped SMTP cost and sender-reputation burn. `/outreach/generate` and `/autohunt/run` (LLM-heavy) are likewise unthrottled.

## Proposed solution
Add `Depends(enforce_expensive_rate_limit)` to the outreach router (and to `/autohunt/run`). Type `to` as `EmailStr`; ideally constrain recipients to a verified contact derived from `application_id` rather than caller-supplied. Add a per-window send quota distinct from the generic limiter.

## Acceptance criteria
- [ ] Rate limiting enforced on `/outreach/send`, `/outreach/generate`, `/autohunt/run`
- [ ] `to` validated as an email address
- [ ] Recipient constrained to (or validated against) the application's contact
- [ ] Tests cover rejection of invalid recipient and over-limit sends

**Estimated effort:** S
**References:** OWASP API4:2023 (Unrestricted Resource Consumption), CWE-770

StevSant:
> **Supersedes #185.** That PR's functional CI was green, but GitGuardian flagged a secret-shaped *unit-test fixture* that survived in an intermediate commit even after being reworded — clearing it required rebuilding the branch history. This branch contains the **byte-identical final tree** rebuilt as three clean per-issue commits (verified: `git diff` between the two heads is empty), so none of the commits contain the flagged fixture. #185 is closed in favor of this PR; its branch is left on origin for comparison.

## Summary

P1 backend security hardening — three High-severity findings from the 2026-07-05 audit, one commit each.

### #128 — LaTeX compiled with `-no-shell-escape` + paranoid file access
`LatexCompiler._run_once` invoked `xelatex` without `-no-shell-escape`, so untrusted LaTeX (raw user CVs and LLM-spliced optimized CVs) could read arbitrary files via `\input{/etc/passwd}` (restricted shell-escape) or achieve RCE via `\write18` (full shell-escape). The fix adds `-no-shell-escape` to every invocation and sets `openin_any=p` / `openout_any=p` in the subprocess env so TeX cannot read or write files outside its temp dir. Both compile paths (original CV and optimized CV) share this single argv, so the argv-level test covers both.

Closes #128

### #129 — Admin password hashed with scrypt, compared in constant time
`AuthService.login` stored the admin password in plaintext and compared with `==` (timing side channel, no defense in depth). Added a stdlib `scrypt`-based hasher (`hiresense.identity.domain.hash_password` / `verify_password`) and a new optional `AUTH_PASSWORD_HASH` setting that takes precedence over `AUTH_PASSWORD`; when a hash is set the plaintext is never retained by the service. When only plaintext is configured (the local-mode ephemeral dev password), both username and password are now compared with `secrets.compare_digest` (as UTF-8 bytes — `compare_digest` raises `TypeError` on non-ASCII `str`, which would otherwise turn a crafted credential into a 500). A fully-blank credential can never authenticate.

Closes #129

### #130 — Rate-limited + recipient-validated `/outreach/send`
The outreach router was gated only by `require_auth` with no rate limiting, and `SendRequest.to` was a bare `str` untied to any real contact — an authenticated open email relay with unbounded LLM/SMTP cost. Fixes:
- `Depends(enforce_expensive_rate_limit)` added to `POST /outreach/send`, `POST /outreach/generate`, and `POST /autohunt/run`.
- `SendRequest.to` is now `EmailStr` (added the `pydantic[email]` extra).
- New config-driven recipient-domain allowlist `OUTREACH_ALLOWED_RECIPIENT_DOMAINS` (empty = allow any valid email). A send to a domain outside the allowlist raises `RecipientNotAllowedError` → HTTP 400, before any email dispatch.

Closes #130

## Decisions taken

- **#128 — no real-compile in CI.** `xelatex` isn't installed on the dev machine or CI, so the load-bearing regression test asserts the argv/env at the subprocess boundary (mocked `subprocess.run`). A real-compile test that feeds `\write18`/`\input{/abs}` and expects `LatexCompileError` is included but guarded with `@pytest.mark.skipif(shutil.which("xelatex") is None)`.
- **#128 — added `openin_any=p`/`openout_any=p`.** `-no-shell-escape` alone stops `\write18` (RCE) but `\input{/etc/passwd}` is a core TeX primitive, not shell-escape; the paranoid file-access env vars are what actually close the file-read surface called out in the finding.
- **#129 — stdlib scrypt, no new dependency.** Used `hashlib.scrypt` + `secrets.compare_digest` (self-describing `scrypt$n$r$p$salt$hash` format) rather than adding passlib/argon2. Cost params (n=16384, r=8, p=1 ≈ 16 MiB) stay under scrypt's default 32 MiB `maxmem` cap.
- **#129 — kept the local-mode flow intact.** `config.mode.apply_mode` still generates an ephemeral plaintext dev password when `AUTH_PASSWORD` is blank in local mode; `AUTH_PASSWORD_HASH` is optional and only takes precedence when set, so the degraded dev boot is unchanged.
- **#130 — recipient constrained via a domain allowlist, not a stored contact.** `TrackedApplication` has no contact/email field, so there is no per-application contact to validate against. The available and honest interpretation of "constrain the recipient" is (a) reject malformed addresses with `EmailStr` and (b) an optional operator-configured domain allowlist. Default (empty) preserves current behaviour.
- **#130 — reused `enforce_expensive_rate_limit`, no separate send quota.** The issue's "consider a distinct per-window send quota" was optional; the existing per-IP sliding-window limiter already bounds send/generate/run cost and satisfies the over-limit test. A separate limiter would add app.state wiring for marginal benefit — deferred.
- **#130 — `RecipientNotAllowedError` subclasses `ValueError`** so existing `except ValueError` handlers still catch it; the route catches it first and maps to 400 (distinct from the 404 used for a missing application).
- **History rebuild (supersedes #185).** GitGuardian scans every commit in a PR, so the softened-later fixture kept failing the check on #185. Rather than force-push rewritten history onto the reviewed branch, the identical final tree was re-committed as three per-issue commits on this fresh branch.

## Verification

Run from the worktree `backend/` with `UV_LINK_MODE=copy`:

- `uv run ruff check .` → **All checks passed!**
- `uv run python -m pytest` → full suite **green** (exit code 0, DB-free SQLite run). New tests:
  - `tests/unit/test_latex_compiler.py` — argv includes `-no-shell-escape`, env has `openin_any/openout_any=p`; skipif real-compile guard.
  - `tests/unit/identity/test_password_hasher.py` + `tests/unit/identity/test_auth.py` — hash verify correct/incorrect, plaintext not retained, wrong-username and blank-credential rejection, non-ASCII credential → clean 401.
  - `tests/unit/outreach/test_send.py` — recipient allowlist reject/allow (case-insensitive).
  - `tests/integration/test_outreach_endpoints.py` — invalid email → 422, over-limit send → 429.
- Head tree verified byte-identical to #185's head (which passed the full backend + frontend CI matrix).

Repo traps checked:
- **`.env.example` updated** for `AUTH_PASSWORD_HASH` and `OUTREACH_ALLOWED_RECIPIENT_DOMAINS` (both with comments).
- **Config group used**, not hardcoded: `auth_password_hash` in `config/groups/core.py`, `outreach_allowed_recipient_domains` in `config/groups/outreach.py`.
- **`__init__` re-exports** added: `hash_password`/`verify_password` in `identity/domain`, `RecipientNotAllowedError` in `outreach/domain`.
- **`registry.py`** — n/a, no new ORM classes.

https://claude.ai/code/session_01PmhxbVCx2ErqBBjibnqHY8

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -38,6 +38,11 @@ SEMANTIC_PROFILE_CACHE_SIZE=8
 #   python -c "import secrets; print(secrets.token_urlsafe(48))"
 AUTH_USERNAME=admin
 AUTH_PASSWORD=changeme
+# Preferred over AUTH_PASSWORD: a scrypt hash so the plaintext never lives in
+# env/config. When set, it takes precedence and AUTH_PASSWORD is ignored.
+# Generate with:
+#   python -c "from hiresense.identity.domain import hash_password; print(hash_password('your-password'))"
+AUTH_PASSWORD_HASH=
 JWT_SECRET_KEY=change-this-to-a-random-secret
 # Role embedded in issued tokens. Single-user instance is admin by default;
 # set to a non-admin value to exercise the admin-only route gate.
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -315,6 +315,10 @@ SMTP_USERNAME=
 SMTP_PASSWORD=
 SMTP_USE_TLS=true
 OUTREACH_FROM_EMAIL=
+# Recipient-domain allowlist for POST /outreach/send (comma/JSON list). Empty
+# allows any valid email; set to constrain sends and reduce the authenticated
+# open-relay/cost surface, e.g. ["example.com","gmail.com"].
+OUTREACH_ALLOWED_RECIPIENT_DOMAINS=[]
 
 # --- Scheduler (Autopilot Phase 1) ---
 # Self-drive the recurring pipeline in-process. Set true on EXACTLY ONE process.
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -6,7 +6,7 @@ requires-python = ">=3.12"
 dependencies = [
     "fastapi>=0.139.0",
     "uvicorn[standard]>=0.50.2",
-    "pydantic>=2.13.4",
+    "pydantic[email]>=2.13.4",
     "pydantic-settings>=2.14.2",
     "sqlalchemy[asyncio]>=2.0.51",
     "asyncpg>=0.30.0",
--- a/backend/src/hiresense/adapters/latex/compiler.py
+++ b/backend/src/hiresense/adapters/latex/compiler.py
@@ -76,11 +76,19 @@ class LatexCompiler:
             return pdf_path.read_bytes()
 
     def _run_once(self, working_dir: Path, source: Path):
+        import os
         import subprocess
 
+        # Untrusted LaTeX reaches this compiler (raw user CVs, LLM-spliced
+        # optimized CVs). Disable shell-escape entirely so `\write18{...}` can
+        # never spawn a subprocess (RCE), and set openin_any/openout_any to
+        # "paranoid" so `\input{/etc/passwd}` and friends cannot read/write
+        # files outside the compilation temp dir.
+        env = {**os.environ, "openin_any": "p", "openout_any": "p"}
         return subprocess.run(
             [
                 self._compiler,
+                "-no-shell-escape",
                 "-interaction=nonstopmode",
                 "-halt-on-error",
                 "-output-directory",
@@ -90,6 +98,7 @@ class LatexCompiler:
             cwd=working_dir,
             capture_output=True,
             timeout=self._timeout,
+            env=env,
         )
 
     @staticmethod
--- a/backend/src/hiresense/autohunt/api/routes.py
+++ b/backend/src/hiresense/autohunt/api/routes.py
@@ -4,12 +4,12 @@ from fastapi import APIRouter, Depends, Query, Response
 
 from hiresense.autohunt.api.dependencies import get_autohunt_service
 from hiresense.autohunt.domain import AutoHuntService, Digest
-from hiresense.identity.api.dependencies import require_auth
+from hiresense.identity.api.dependencies import enforce_expensive_rate_limit, require_auth
 
 router = APIRouter(prefix="/autohunt", tags=["autohunt"], dependencies=[Depends(require_auth)])
 
 
-@router.post("/run", response_model=Digest)
+@router.post("/run", response_model=Digest, dependencies=[Depends(enforce_expensive_rate_limit)])
 async def run(service: AutoHuntService = Depends(get_autohunt_service)) -> Digest:
     return await service.run()
 
--- a/backend/src/hiresense/bootstrap/identity.py
+++ b/backend/src/hiresense/bootstrap/identity.py
@@ -10,4 +10,5 @@ def build_identity(settings: Settings) -> IdentityProvider:
         password=settings.auth_password,
         jwt_secret=settings.jwt_secret_key,
         role=settings.auth_role,
+        password_hash=settings.auth_password_hash,
     )
--- a/backend/src/hiresense/bootstrap/outreach.py
+++ b/backend/src/hiresense/bootstrap/outreach.py
@@ -45,5 +45,6 @@ def build_outreach(
         language=s.default_language,
         portfolio_citation=portfolio_citation,
         sender=sender,
+        allowed_recipient_domains=tuple(s.outreach_allowed_recipient_domains),
     )
     return OutreachBuild(provider=OutreachProvider(outreach_service=service), service=service)
--- a/backend/src/hiresense/config/groups/core.py
+++ b/backend/src/hiresense/config/groups/core.py
@@ -37,6 +37,11 @@ class CoreSettings(BaseSettings):
     # creds, see config.mode.apply_mode); required in production.
     auth_username: str = ""
     auth_password: str = ""
+    # Optional scrypt hash of the admin password (scrypt$n$r$p$salt$hash). When
+    # set it takes precedence over AUTH_PASSWORD and the plaintext is never
+    # retained by the auth service. Generate with:
+    #   python -c "from hiresense.identity.domain import hash_password; print(hash_password('pw'))"
+    auth_password_hash: str = ""
     jwt_secret_key: str = ""
 
     @field_validator("app_mode", mode="before")
--- a/backend/src/hiresense/config/groups/outreach.py
+++ b/backend/src/hiresense/config/groups/outreach.py
@@ -22,6 +22,10 @@ class OutreachSettings(BaseSettings):
     smtp_password: str = ""
     smtp_use_tls: bool = True
     outreach_from_email: str = ""
+    # Recipient-domain allowlist for POST /outreach/send. Empty (default) allows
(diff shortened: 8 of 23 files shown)


StevSant:
## Self code-review (full diff `origin/main...HEAD`)

Reviewed all 24 files for bugs, security gaps, and missed acceptance criteria. One issue was found during review and is already incorporated in the #129 commit; the rest verified clean. (This review was originally performed on #185; this branch carries the byte-identical tree.)

### Found & fixed
- **[HIGH] Non-ASCII credential → 500 (#129).** `secrets.compare_digest` raises `TypeError` when given `str` args containing non-ASCII characters. A crafted non-ASCII username/password would have propagated as an unhandled 500 instead of a clean 401. Fixed by comparing UTF-8-encoded bytes (`username.encode("utf-8")` …), with regression test `test_login_handles_non_ascii_input_without_error`.

### Verified clean
- **#128 order of flags.** `-no-shell-escape` precedes `-interaction`/`-halt-on-error`; xelatex accepts it. `env` merges `os.environ` so PATH etc. survive; `openin_any=p`/`openout_any=p` are additive.
- **#129 local-mode boot.** `AuthService` only nulls the plaintext when a hash is present, so the ephemeral dev password path (blank `AUTH_PASSWORD` → generated plaintext) is unchanged. Blank-credential guard returns False before any compare.
- **#129 no plaintext retention.** When `AUTH_PASSWORD_HASH` is set, `self._password` is `""` — asserted by `test_password_hash_takes_precedence_and_plaintext_not_retained`.
- **#130 exception ordering.** `RecipientNotAllowedError` subclasses `ValueError`, and the route catches it *before* the generic `ValueError` arm (→ 400 vs 404). The missing-application check runs first in the service, so a missing app still 404s.
- **#130 allowlist default.** Empty `OUTREACH_ALLOWED_RECIPIENT_DOMAINS` short-circuits to allow-any, preserving existing behaviour; comparison is lowercased on both sides (case-insensitive), verified by test.
- **#130 EmailStr.** `pydantic[email]` extra added to `pyproject.toml` + `uv.lock`; malformed recipient → 422 (Pydantic), verified by integration test.

### Notes / non-blocking
- The recipient allowlist is domain-level, not a per-application contact (see PR "Decisions taken" — `TrackedApplication` stores no contact address). If a contact field is added later, tightening this to an exact-address match is the natural follow-up.
- No new ORM classes, so `infrastructure/registry.py` needs no change.

StevSant:
Superseded by the v3 PR. The GitGuardian incident (Generic Password) was the fixture literal "s3cret" in tests/unit/identity/test_password_hasher.py — it was still present in this branch's final tree, so the rebuild here couldn't clear it. v3 replaces the literal with a plain-English passphrase in every commit. Branch left on origin; safe to delete after v3 merges.

PR #196, opened 2026-07-20
fix(identity): auth hardening — login limiter, limiter eviction, JWT claims, audit actor

### LINKED ISSUE #138: Bug: Admin audit actor stores the full JWT payload dict instead of the username
**Severity:** Medium
**Location:** `backend/src/hiresense/admin/api/routes_llm_settings.py:47,106,127`; `identity/api/dependencies.py:55-64`
**Confidence:** High

## Description
`require_admin` returns the full decoded token `payload` (a dict). The LLM-settings routes inject it as `actor: Annotated[str, Depends(require_admin)]` and pass `actor=actor` into `LLMSettingsService`, which stores it as `updated_by=actor` (a String column). So the audit field is written the whole payload dict (stringified or a coercion error) rather than the admin username — the audit trail is corrupted and the `str` annotation is inaccurate.

## Proposed solution
Add a small dependency returning `payload["sub"]` for the `actor` parameter, or pass `actor=payload["sub"]` at call sites. Keep `require_admin` as the gate.

## Acceptance criteria
- [ ] `updated_by` records the username, not the payload dict
- [ ] Test asserts the stored actor equals the token subject

**Estimated effort:** XS

### LINKED ISSUE #141: Security: Add a dedicated, stricter login rate limiter
**Severity:** Medium
**Location:** `backend/src/hiresense/identity/api/routes.py:20-32`; limits in `config/groups/core.py:75-77`
**Confidence:** Medium

## Description
Login is gated by the shared `enforce_expensive_rate_limit` (default 30 req/60s per IP), allowing ~1800 guesses/hour against the single admin credential, and the counter is shared with all other "expensive" endpoints so auth and ingestion/matching traffic contend for one budget. There is no separate limiter or lockout for authentication. Combined with a weak-if-guessable password this is a viable online brute-force surface.

## Proposed solution
Add a dedicated, tighter login limiter (e.g. 5–10 attempts / 15 min per IP, ideally keyed on username too), independent of the expensive-op bucket; optionally exponential backoff after repeated failures.

## Acceptance criteria
- [ ] Login has its own stricter limiter separate from the expensive bucket
- [ ] Test asserts login lockout triggers before the generic limit

**Estimated effort:** S
**References:** OWASP ASVS V2.2.1, CWE-307

### LINKED ISSUE #152: Performance: Rate limiter never evicts idle keys (unbounded memory growth)
**Severity:** Low
**Location:** `backend/src/hiresense/kernel/rate_limit.py:19,25-36`
**Confidence:** High

## Description
`_events` maps every client-IP key to a deque. Expired timestamps are popped, but an emptied deque and its key are never removed, and idle keys are never reclaimed. Over a long-lived process facing many distinct IPs the dict grows without bound — a slow memory leak. Keying is on the real socket peer (not spoofable), so exploitation needs many real IPs, but organic growth over uptime is real.

## Proposed solution
`del self._events[key]` when the deque is empty after pruning; or periodically sweep stale entries.

## Acceptance criteria
- [ ] Emptied keys removed
- [ ] Test asserts the map shrinks after a window elapses

**Estimated effort:** XS

### LINKED ISSUE #154: Security: Enforce iss/aud claims (and require exp) on JWT validation
**Severity:** Low
**Location:** `backend/src/hiresense/identity/domain/services.py:21-33`
**Confidence:** High

## Description
Tokens are HS256 with only `sub`/`role`/`exp`. `validate_token` verifies signature+algorithm (pinned to HS256, so no alg-confusion) and expiry, but sets no `iss`/`aud`, has no `jti`/revocation, and a fixed 24h lifetime with no refresh/logout. A leaked token is valid a full day with no revocation, and any service reusing the secret would accept the same token shape.

## Proposed solution
Add and enforce `iss`/`aud`; require `exp` explicitly via decode `options`; consider a shorter lifetime; add `jti` + denylist if logout/revocation is needed.

## Acceptance criteria
- [ ] `iss`/`aud` set and validated
- [ ] `exp` required in decode options
- [ ] Tests cover wrong-issuer/audience rejection

**Estimated effort:** S
**References:** RFC 8725 JWT BCP

StevSant:
Batch of four auth/identity hardening fixes. One commit per issue, scoped by module.

### #152 — Rate limiter never evicts idle keys (`kernel`)
`SlidingWindowRateLimiter._events` grew one deque per distinct client IP forever: idle keys are never revisited, so their deques are never pruned. Added a throttled (once-per-window) `O(keys)` sweep that drops keys whose newest timestamp predates the window, so the map stays bounded. Still-active keys (with only some stale entries) are kept.

### #154 — Enforce iss/aud/exp on JWT validation (`identity`)
`validate_token` verified only signature + expiry. Now issues and validates `iss`/`aud` and **requires** `exp`/`iss`/`aud` to be present. Discovered that python-jose's `verify_*` only checks a claim it *finds* — a token omitting `aud`/`iss` would otherwise pass — so `require_exp`/`require_iss`/`require_aud` are all set. Issuer/audience default to stable identifiers (`hiresense` / `hiresense-api`), issued and validated by the same service, so the local-mode ephemeral-secret dev flow keeps working.

### #141 — Dedicated, stricter login rate limiter (`identity`)
`POST /auth/login` shared the expensive-ops bucket (30 req/60s), so auth contended with ingestion/matching traffic and allowed ~1800 guesses/hr. Added a separate `login_rate_limiter` (default **5 attempts / 15 min** per IP, OWASP ASVS V2.2.1 / CWE-307) and pointed the login route at it exclusively, removing login from the shared bucket.

### #138 — Audit actor stored the full JWT payload dict (`admin`, `identity`)
`require_admin` returns the whole decoded payload; the LLM-settings routes injected it as `actor: str` and wrote it to `updated_by`, corrupting the audit trail. Added a `require_admin_actor` dependency (depends on `require_admin` so the gate still applies; returns `payload["sub"]`) and switched the three `actor` call sites to it. The router-level `require_admin` gate is unchanged.

## Closes
Closes #152
Closes #154
Closes #141
Closes #138

## Decisions taken
- **#141 keying & placement:** keyed on client IP (not username), matching the existing non-spoofable socket-peer keying — for a single-admin instance this is effectively username-scoped anyway. Replaced (not stacked on) the expensive-bucket dependency on `/auth/login` so auth no longer shares a budget with other traffic. Defaults 5 / 900s.
- **#154 defaults & presence:** `iss`/`aud` are non-secret identifiers, so stable defaults are safe and keep local mode working. Went beyond "set and validate" to also **require presence** after finding jose skips absent claims.
- **#152 approach:** combined the issue's two options into a throttled idle-key sweep — `del`-on-empty alone can't reclaim keys that are never revisited (the actual leak).
- **#138 shape:** one reusable `require_admin_actor` dependency rather than `payload["sub"]` at each call site.

## Verification
- `uv run ruff check .` → **All checks passed**
- Focused: `pytest tests/unit/test_rate_limit.py tests/unit/identity/ tests/unit/admin/ tests/integration/test_login_rate_limit.py` → **81 passed**
- Full suite: `uv run python -m pytest` → **1376 passed**, 1 failure = `test_settings_autopilot::test_autopilot_pipeline_settings_defaults`, a pre-existing *environmental* failure (local `.env` sets `AUTOPILOT_PIPELINE_ENABLED=true`; green in CI — documented). `test_public_symbols_importable` is a pre-existing collection-order flake (passes in isolation; only asserts config flat-access fields untouched here) — not introduced by this PR.

## Repo-trap checks
- **Config in groups + `.env.example`:** `jwt_issuer`, `jwt_audience`, `login_rate_limit_*` added to `config/groups/core.py` **and** `.env.example` with comments — no hardcoded thresholds.
- **Package `__init__` re-exports:** `require_admin_actor` re-exported from `identity/api/__init__.py` and `admin/api/dependencies.py` `__all__`.
- **ORM registry / migration:** N/A — no schema or ORM changes.

> Note: `.env.example` and `config/groups/core.py` are also touched by sibling PRs in this batch; expect trivial conflict resolution at merge (additive blocks only).


StevSant:
## Self-review (critical pass over the full diff)

No CRITICAL or HIGH issues found. Correctness of the JWT claim enforcement was verified with a standalone jose smoke test (valid token accepted; wrong-iss, wrong-aud, missing-exp, and missing-iss/aud all rejected). Notes and residual considerations:

**#154 — `identity/domain/services.py:60-88`**
- `options={"require_exp": True, "require_iss": True, "require_aud": True}` is load-bearing: python-jose's `verify_*` only checks claims it *finds*, so without the `require_*` flags a token that simply omits `aud`/`iss` passes. Confirmed empirically before finalizing.
- Residual (documented, not fixed): a second deployment reusing the same `JWT_SECRET_KEY` **and** the default `iss`/`aud` would still cross-accept tokens. The `.env.example` comment tells operators to override `JWT_ISSUER`/`JWT_AUDIENCE` per deployment. Full isolation is opt-in, which is the right default for a single self-hosted instance.
- Existing tokens minted before deploy (without `iss/aud`) become invalid → a re-login. Acceptable on deploy; the app issues and validates consistently from here on.

**#141 — `identity/api/dependencies.py:66-96`, `identity/api/routes.py:23-33`, `main.py`**
- The login limiter is checked before authentication, so **successful** logins also consume the budget. 5/15 min is generous for a single admin, but a human fat-fingering the password a few times then succeeding still has headroom. Fine for this app.
- Keying is `request.client.host` (same as the pre-existing expensive limiter). Behind a reverse proxy with no `X-Forwarded-For` handling this collapses to the proxy IP — a pre-existing property of the limiter design, out of scope here.
- `_enforce_rate_limit` helper de-duplicates the two enforcers cleanly; the `None`-limiter no-op path is preserved (bare test apps / disabled config).

**#152 — `kernel/rate_limit.py:38-60`**
- The sweep runs inside `self._lock`, so the comprehension + `del`s are atomic w.r.t. concurrent `allow()`. Throttled to once per window → amortized `O(keys)`. Between sweeps, memory is bounded by (distinct IPs seen in one window), not unbounded — the leak is closed. `_last_sweep` is seeded at construction so the first sweep only fires after a full window.
- Gating on `events[-1] < cutoff` (newest timestamp) avoids per-deque pruning during the sweep and correctly keeps still-active keys.

**#138 — `identity/api/dependencies.py:113-124`, `admin/api/routes_llm_settings.py:47,109,131`**
- `require_admin_actor` depends on `require_admin`, so the admin gate is unchanged and only the return value narrows to `sub`. `payload["sub"]` could in theory `KeyError`, but every token this service issues carries `sub` and validation now enforces the token shape; a `.get` would only mask a malformed-token bug. Left as a direct index deliberately.
- Router-level `dependencies=[Depends(require_admin)]` is untouched — the gate and the actor extraction are independent.

**Conventions:** config lives in `config/groups/core.py` + `.env.example` (no hardcoded thresholds); `require_admin_actor` is re-exported from both package `__init__`/`__all__`; no ORM/schema change so no migration. Ruff clean; full suite green except the two pre-existing/environmental items called out in the PR body.

StevSant (commented):
## Code review — #196

**Done well**
- `backend/src/hiresense/kernel/rate_limit.py` — the idle-key eviction condition (`not events or events[-1] < cutoff`) exactly mirrors the per-key staleness check used inside `allow()`, so a key is only evicted when *all* of its events are stale — a still-active key (partially stale deque) is correctly retained. The sweep is throttled to once per window and runs under the existing lock, so no new race window is introduced.
- `backend/src/hiresense/identity/domain/services.py` — correctly identifies and works around the real python-jose gotcha: `verify_aud`/`verify_iss` only validate a claim if it's *present*, so a forged token that simply omits `aud`/`iss` would otherwise sail through. Setting `require_exp`/`require_iss`/`require_aud` closes that gap, and there's a test (`test_validate_rejects_token_missing_exp`) that hand-crafts exactly such a token to prove it.
- `backend/src/hiresense/identity/api/dependencies.py` — good refactor extracting `_enforce_rate_limit` so the login and expensive limiters share one implementation instead of duplicating the 429/Retry-After logic.
- `backend/src/hiresense/admin/api/dependencies.py` / `routes_llm_settings.py` — `require_admin_actor` depends on `require_admin` (not a standalone check), so the admin gate still applies; FastAPI's per-request dependency caching means it isn't evaluated twice even though the router also declares `dependencies=[Depends(require_admin)]`. All three `actor` call sites (`update_settings`, `upsert_override`, `clear_override`) were switched consistently.
- Config additions (`jwt_issuer`, `jwt_audience`, `login_rate_limit_*`) follow the repo convention: added to `config/groups/core.py` with comments and mirrored in `.env.example`, no hardcoded values in application code.
- `require_admin_actor` and the login-limiter dependency are re-exported through the relevant package `__init__.py`/`__all__`, per the "import from the contextual package" convention.
- Test coverage is solid and targeted: issuer/audience mismatch, missing-claim rejection, sweep eviction vs. active-key retention, and an integration test proving the login limiter trips independently of (and before) the shared expensive bucket.

- 🟡 `backend/src/hiresense/identity/domain/services.py:60-75` — Adding `require_iss`/`require_aud`/`require_exp` on `validate_token` will reject any session token issued by the pre-PR code (which had no `iss`/`aud` claims) the moment this deploys, forcing every logged-in user to re-authenticate. For this single-admin, self-hosted app that's a minor/acceptable one-time inconvenience (worst case a 401 + re-login), not a functional bug — flagging only so it's an expected/communicated side effect of the rollout rather than a surprise.
- 🟡 `backend/src/hiresense/identity/api/dependencies.py` — `enforce_login_rate_limit` keys on `request.client.host`, identical to the existing expensive-bucket keying. That's consistent with the established (non-spoofable, direct-socket-peer) convention and is explicitly called out as a deliberate decision in the PR body, so not a new gap — just noting it inherits the same "no reverse-proxy XFF trust" assumption the rest of the app already makes.

**Verified**
- Full diff read with 40 lines of context for every changed file: `kernel/rate_limit.py`, `identity/domain/services.py`, `identity/api/{dependencies,provider,routes,__init__}.py`, `admin/api/{dependencies,routes_llm_settings}.py`, `bootstrap/identity.py`, `main.py`, `config/groups/core.py`, `.env.example`, and all four new/modified test files.
- JWT validation is the single code path (`_authenticate` → `auth_service.validate_token`) for both the `Authorization: Bearer` header and the httpOnly session cookie (PR #186's flow) — no second/refresh verification path exists in the codebase (grepped for `refresh` and other `validate_token` call sites), so `iss`/`aud`/`exp` enforcement covers every entry point.
- `jwt_issuer`/`jwt_audience` are untouched by `config/mode.py`'s local-mode degradation logic (only `jwt_secret_key` gets an ephemeral per-boot value there) — they always use the stable defaults, so the `APP_MODE=local` ephemeral-secret boot path is unaffected and keeps working.
- Rate-limiter sweep logic: confirmed no `RuntimeError: dict changed size during iteration` risk (stale keys collected into a list before deletion) and no scenario where an actively-throttled key's remaining quota is wiped by the sweep.
- `require_admin_actor` returns `payload["sub"]` (a string), not the full payload dict, fixing the audit-trail corruption described in #138.
- Cross-checked `test_public_symbols_importable` (the reported pre-existing flake) — it only asserts flat-access on `otel_enabled`, `database_url`, `weight_semantic`, `portfolio_ref_prefix`, `app_mode`, none of which this PR touches, supporting the PR's claim that the failure is unrelated/environmental.
- No other module reads `SlidingWindowRateLimiter._events` externally (grepped `_events` under `backend/src`), so the new eviction behavior can't break an unrelated consumer.

**Verdict:** LGTM ✅

Review by Claude Code

Issue #133, opened 2026-07-06
Bug: Admin CSV export uses window.open and bypasses the auth interceptor (always 401)


StevSant:
**Severity:** Medium
**Location:** `frontend/src/app/pages/admin/admin-usage.component.ts:157-168`; `core/services/admin-usage.service.ts:49-55`; backend gate `admin/api/routes_usage.py:20-24`
**Confidence:** High

## Description
`exportCsv()` builds the export URL and calls `window.open(url, '_blank')`. The JWT is attached only by `authInterceptor`, which runs on Angular `HttpClient` requests — not on a plain browser navigation. The endpoint mounts `Depends(require_admin)`, so the navigation arrives with no `Authorization` header and returns 401; the CSV never downloads. The method's own comment describes the correct blob-download approach but the code does the opposite. The correct pattern already exists (`apply-tab.component.ts` `downloadBundle`, `cv-tab.component.ts` `triggerDownload`).

## Proposed solution
Fetch via `HttpClient` with `responseType: 'blob'` (interceptor attaches the token), then download via `URL.createObjectURL` + a temporary anchor, mirroring `apply-tab`.

## Acceptance criteria
- [ ] CSV downloads successfully as an authenticated admin
- [ ] Spec covers the blob-download path

**Estimated effort:** S

StevSant:
Verified against current `main` (post-#186 httpOnly-cookie migration) — this no longer reproduces, closing as already fixed.

The report was based on the JWT bearer header + `authInterceptor`, both of which the cookie migration removed (commit c03b8d8, PR #131/#186). Session auth is now a same-origin httpOnly cookie, and the export works end to end:

- `frontend/src/app/pages/admin/admin-usage.component.ts:157-168` — `exportCsv()` still uses `window.open`, but the migration rewrote its comment to reflect the new mechanism: a same-origin top-level navigation carries the httpOnly session cookie automatically (same-site even under SameSite=Strict), so no interceptor is needed.
- `frontend/src/environments/environment.ts` — `apiUrl` is the relative `/api`, so the export URL built in `core/services/admin-usage.service.ts:49-55` (`/api/admin/usage/export?...`) is same-origin with the app; the cookie rides the navigation.
- `backend/src/hiresense/admin/api/routes_usage.py:103-120` — the endpoint returns a `StreamingResponse` with `Content-Disposition: attachment; filename=llm_usage.csv`, so the browser downloads the file rather than rendering it inline.

[...]

PR #198, opened 2026-07-20
fix(frontend): security hardening — CSP, URL scheme validation

### LINKED ISSUE #153: Security: Add a Content-Security-Policy

### LINKED ISSUE #169: Security: Validate URL scheme before window.open on ingested job URLs
**Severity:** Low
**Location:** `frontend/src/app/pages/applications/components/apply-tab.component.ts:129-135`
**Confidence:** Medium

## Description
`aggregate().url` (ingested, attacker-influenceable) is passed straight to `window.open(url, '_blank', 'noopener')`. Unlike `[href]` bindings (which Angular sanitizes), `window.open` performs no sanitization, so a `javascript:` URL in a malicious posting would execute. Impact is limited by `noopener`/new-window origin, hence Low, but it's a genuine unsanitized sink.

## Proposed solution
Validate the scheme before opening (allow only `http:`/`https:`), e.g. `const u = new URL(url); if (u.protocol==='http:'||u.protocol==='https:') window.open(...)`. Optionally centralize as `openExternal(url)`.

## Acceptance criteria
- [ ] Non-http(s) schemes rejected before `window.open`
- [ ] Shared helper (optional) covers other `window.open` sites

**Estimated effort:** XS

StevSant:
Frontend security-hardening batch. Two of the four assigned findings were already resolved by the httpOnly-cookie migration (#186) and are closed as stale; the other two are implemented here.

### #153 — Content-Security-Policy (`fix(frontend)`)
The SPA had no CSP. Added one as a response header from the nginx server that serves the built app (`frontend/nginx.conf`) — the backend security-headers middleware only covers `/api` JSON responses, not the document. `script-src` is kept strict at `'self'`; every relaxation is tied to a real sink found in the code:

| Directive | Value | Why |
|---|---|---|
| `script-src` | `'self'` | only the external Angular module bundle; no inline scripts/handlers |
| `style-src` | `'self' 'unsafe-inline' https://fonts.googleapis.com` | Angular injects component styles as inline `<style>` at runtime (no nonce wired); Google Fonts stylesheet |
| `font-src` | `'self' https://fonts.gstatic.com` | Google Fonts woff2 files |
| `img-src` | `'self' data: https:` | external company logos (company-intel) come from arbitrary domains |
| `connect-src` | `'self'` | app only calls same-origin `/api` |
| `frame-src` | `'self' blob:` | the CV PDF-preview iframe loads a `blob:` URL |
| `object-src` / `base-uri` / `form-action` / `frame-ancestors` | `'none'` / `'self'` / `'self'` / `'none'` | hardening |

To keep `script-src` free of `'unsafe-inline'`, the production build's critical-CSS inlining is disabled (`angular.json` → `optimization.styles.inlineCritical: false`). Beasties otherwise emits an inline `onload="this.media='all'"` handler on the stylesheet link, which a strict `script-src` blocks — leaving the app unstyled. The main stylesheet now loads as a normal render-blocking `<link>` (cost: ~8 kB CSS, one extra round trip before FCP).

### #169 — URL-scheme validation before `window.open` (`fix(applications)`)
`apply-tab`'s "open job & mark applied" passed the ingested (attacker-influenceable) `aggregate().url` straight to `window.open`, which performs no sanitization — a `javascript:`/`data:` URL in a malicious posting would execute in the app origin. Added a shared `openExternalUrl()` helper (`core/utils/open-external-url.ts`) that opens only `http(s)` URLs and rejects unparseable ones, with a spec covering the accept/reject paths.

Closes #169
Closes #153

## Closed as stale (already fixed by #186, not by this PR)

- **#133** — Admin CSV export "always 401". The bearer header + `authInterceptor` the report was based on are gone. The export URL is same-origin (`environment.apiUrl` = `/api`), so the httpOnly session cookie rides the `window.open` navigation automatically, and the backend returns `Content-Disposition: attachment` — the CSV downloads. Closed with file:line evidence.
- **#156** — `authInterceptor` sends bearer token to all origins. `auth.interceptor.ts` was deleted; `credentialsInterceptor` sets `withCredentials` only for the API base, and its spec already asserts a third-party URL gets no credentials — matching the issue's acceptance criteria verbatim. Closed with evidence.

## Decisions taken

- **CSP location:** served only via the nginx header, not an `index.html` `<meta>`. A strict `<meta>` CSP would also apply to `ng serve`, breaking dev HMR (eval + ws). The header is the issue's preferred option and satisfies "CSP served (header or meta)".
- **`inlineCritical: false`:** chosen over adding `'unsafe-inline'`/`'unsafe-hashes'` to `script-src`. Keeping `script-src` pristine preserves the CSP's primary XSS protection; the FCP cost is negligible for this app.
- **Admin export left on `window.open`:** the other `window.open` site targets a trusted, app-built same-origin **relative** URL (`/api/...`), not attacker-influenceable, and `new URL()` scheme-parsing would break a relative URL. So the `openExternalUrl` helper is applied only to the ingested job URL (#169's exact target).

## Verification

- `npm run lint` — clean (0 errors, 0 warnings).
- `npm test` — 80 files, 453 tests passed (includes the new `open-external-url.spec.ts`).
- `npm run build` — production build succeeds; built `index.html` confirmed to contain **no** inline scripts or event handlers (so `script-src 'self'` holds).
- **CSP runtime smoke test:** served the built `dist` with the exact policy header and loaded it headless — the app booted, routed to `/login`, and rendered fully styled with **zero CSP-refusal messages**; the only console error was the expected `/api/auth/me` failure (no backend), which confirms `connect-src 'self'` allows the `/api` call. Authenticated-only sinks (`img-src` logos, `frame-src blob:` PDF preview) are covered by the code inventory; a full docker (nginx + backend) pass is recommended as a final check.

## Repo-trap checks

- `ng lint` run explicitly (CI gate that `npm test`/`build` skip) — clean.
- No hardcoded URLs — CSP hosts are the documented Google Fonts endpoints; app URLs use `environment.apiUrl`.
- New util + spec follow conventions (one symbol per file, imported by path); no `.model.ts` needed (no new interfaces).
- Sibling batch PRs may produce trivial conflicts (e.g. shared config files); none expected in these paths.


StevSant:
## Self-review (critical)

Reviewed my own diff for correctness, security, and convention gaps.

**Correctness / maintainability**

- `frontend/nginx.conf` — the CSP is set with a **server-level** `add_header`, which nginx only inherits into a `location` block that defines *no* `add_header` of its own. Both current locations qualify, so it works today, but this is an nginx footgun: the first time someone adds any `add_header` to `location /` or `location /api/`, the CSP silently disappears from that block. I documented it inline; a future contributor must re-declare or use an `include`. Flagging so it's a conscious trade-off, not a surprise.
- `frontend/nginx.conf` — the header is also emitted on `/api/` proxied JSON responses (server-scope inheritance). Harmless (browsers ignore CSP on non-document responses) but slightly noisy; scoping to `location /` was the alternative, at the cost of not covering nginx error pages.
- `core/utils/open-external-url.ts:19` — `openExternalUrl` returns a boolean, but `apply-tab.openJobAndMarkApplied` ignores it and calls `markApplied()` unconditionally. That preserves prior behavior (a rejected/blocked URL still marks the job applied), which I judged correct for the user's intent — but it means a `javascript:`-URL job is silently marked applied with no tab opening and no feedback. Acceptable; noting it.

**Security**

- `img-src 'self' data: https:` is the loosest directive — it permits images from *any* https origin (needed because company logos come from arbitrary domains). Residual risk is low (pixel/cache-timing, not code exec); pinning to known logo hosts isn't feasible since logos are per-company. Deliberate.
- `style-src ... https://fonts.googleapis.com` is technically unused in prod (the Google Fonts stylesheet is inlined by `optimization.fonts`), so it slightly over-permits stylesheet loading from googleapis. Kept for robustness if font inlining is ever disabled; could be dropped for a tighter policy.
- `script-src 'self'` is the load-bearing protection and is kept free of `'unsafe-inline'`/`'unsafe-hashes'` — verified against the actual built `index.html` (no inline scripts or event handlers) and a runtime CSP smoke test (zero refusals).

**Conventions** — new util + spec follow the one-symbol-per-file / import-by-path pattern used across `core/utils`; no `.model.ts` needed. `apply-tab.component.ts` stays CRLF (pre-existing); the edit is a clean 4-line diff with no line-ending churn.

StevSant (commented):
## Code review — #198

**Done well**
- `frontend/src/app/core/utils/open-external-url.ts` — clean, single-purpose helper (one function per file, matches `core/utils/` convention), uses `new URL()` for scheme validation rather than a regex/string-prefix check, which is the robust approach: it correctly normalizes case, and strips embedded tab/newline characters per the WHATWG URL spec, so obfuscated bypass attempts like `java\tscript:alert(1)`, `JAVASCRIPT:alert(1)`, and leading-whitespace `  javascript:alert(1)` all still resolve to `protocol === 'javascript:'` and get rejected (verified locally in Node). `%20javascript:...` fails to parse at all and is caught by the `try/catch`. Good defense against the scheme-obfuscation bypasses called out in the assignment.
- CSP directives are tightly scoped to actual sinks in the code, not just copy-pasted boilerplate: `img-src https:` (company-intel logos are arbitrary domains), `frame-src blob:` (CV PDF preview iframe), `style-src` Google Fonts host — each is justified by a real, greppable usage, and `script-src` stays free of `'unsafe-inline'`/`'unsafe-eval'`.
- The `inlineCritical: false` build-config change to keep `script-src` clean (avoiding Beasties' `onload="this.media='all'"` inline-handler pattern) is a real, correct trade-off, and it's explained in both the PR description and a code comment in `nginx.conf`.
- `nginx.conf` is confirmed to be the file actually deployed: `frontend/Dockerfile` copies it to `/etc/nginx/conf.d/default.conf`, and `docker-compose.yml`'s `frontend` service builds from `./frontend`, so the header lands on the real served path.
- Verified no other required scheme-validation sinks were missed: every other `[href]="…"` binding across the app (job/company/portfolio/profile links) goes through Angular's built-in `DomSanitizer` (`SecurityContext.URL`), which already strips `javascript:` from property bindings — no `bypassSecurityTrustUrl` is used anywhere on attacker-influenceable data (the only `bypassSecurityTrustResourceUrl` call, in `cv-tab.component.ts`, wraps a same-app-created `blob:` URL from a fetched PDF `Blob`, not attacker input). The other `window.open` call (`admin-usage.component.ts:168`) uses an app-built, same-origin, relative `/api/...` URL, not ingested data — correctly left alone as the PR states.
- `connect-src 'self'` is sufficient: grepped for `EventSource`/`WebSocket` and found none in the frontend — no SSE/websocket origin needs to be allow-listed (consistent with earlier audit notes that streaming isn't wired up yet).

**Findings**
- No findings.

**Verified**
- Full diff of all 5 changed files (`git diff origin/main...origin/fix/frontend-security-batch -U40`) and the two new files' full contents via `git show`.
- `frontend/nginx.conf` CSP header is set at server scope (`always`), applies to both `location` blocks, and doesn't conflict with any existing `<meta http-equiv>` CSP in `index.html` (none present) or backend header middleware (no `Content-Security-Policy` set there, confirmed via grep).
- `frontend/Dockerfile` + `docker-compose.yml` wiring confirms `nginx.conf` is the file actually served in the built image.
- Scheme-bypass techniques (tab/newline-embedded scheme, mixed case, leading whitespace, `%20`-prefixed) against `new URL(rawUrl).protocol` tested directly in Node — all correctly rejected by `openExternalUrl`.
- Searched all `window.open` and `[href]` sinks across `frontend/src/app` to confirm the fix's scope (apply-tab's ingested job URL) is the only sink needing code-level scheme validation; everything else is either Angular-sanitized `[href]`, a same-origin app-built URL, or a locally-created `blob:` URL.
- No hardcoded values introduced beyond the documented Google Fonts hosts (matches PR's own repo-trap checklist).

**Verdict:** LGTM ✅

Review by Claude Code

PR #199, opened 2026-07-20
fix(ingestion): SSRF hardening and ingestion performance batch

### LINKED ISSUE #132: Performance: /ingestion/jobs loads the entire corpus and rewrites all scores on every GET
**Severity:** Medium
**Location:** `backend/src/hiresense/ingestion/api/routes.py:271-319`; `ingestion/infrastructure/jobs_repository.py:297-323` (`list_all`/`list_filtered` have no LIMIT)
**Confidence:** High

## Description
`list_jobs` calls `list_all`/`list_filtered` (no LIMIT — the whole bucket is pulled into memory per page view), computes skill scores for all rows, ANN-reranks all rows, and when the profile has skills calls `persist_scores_batch` with a `ScoreUpdate` for the **entire corpus** on every GET — a full-table UPDATE even on plain page-2 pagination. The frontend fires several concurrent `list_jobs` calls, so this is N concurrent full-corpus writes per view. This is a distinct angle from closed #76 (which addressed the re-fetch/re-score trigger on filter/sort): the concern here is write amplification and full-corpus materialization on the read path, scaling O(N) with corpus size.

## Proposed solution
Persist scores only for rows whose score actually changed (diff against loaded value) or only the visible page; push pagination/sort into SQL so the full corpus isn't materialized every request; consider recomputing/persisting scores out-of-band rather than on the read path.

## Acceptance criteria
- [ ] `list_jobs` no longer issues a full-corpus UPDATE per request
- [ ] Result set is bounded (SQL LIMIT/offset or cursor)
- [ ] Benchmark or test demonstrating write volume is proportional to changed rows, not corpus size

**Estimated effort:** L

### LINKED ISSUE #134: Security: SSRF in job-revalidation URL probe (private-IP/metadata access, unbounded body)
**Severity:** Medium
**Location:** `backend/src/hiresense/ingestion/domain/job_revalidation_service.py:163-182`
**Confidence:** Medium

## Description
`_probe` issues `self._http.get(probe_url, follow_redirects=True, ...)` where `probe_url` derives from `job.url` — data ingested from external boards, HN free-text, or CSV import. There is no scheme allowlist and no block on private/link-local ranges (127/8, 10/8, 169.254.169.254, etc.), and redirects are followed, so an allowlisted host can bounce to an internal target. An attacker who gets a crafted listing ingested can make the server issue internal GETs (port scanning, state-changing internal GET endpoints, cloud metadata). `resp.text` is read with no size cap (memory-DoS). The sweep runs in a BackgroundTask after `/fetch` and via `/revalidate`.

## Proposed solution
Validate the resolved URL before probing: http/https only; resolve the hostname and reject private/loopback/link-local/reserved addresses (re-validate after each redirect, or disable `follow_redirects` and cap hops manually). Cap the response read size. Centralize as an SSRF-safe fetch helper reused by all outbound adapters.

## Acceptance criteria
- [ ] Private/loopback/link-local/metadata targets rejected pre-request and post-redirect
- [ ] Response read size capped
- [ ] Tests cover a private-IP URL and a redirect-to-private-IP

**Estimated effort:** M
**References:** OWASP A10:2021 SSRF, CWE-918

### LINKED ISSUE #157: Performance: Wrap sync DB reads in to_thread in analyze_job/get_job routes
**Severity:** Low
**Location:** `backend/src/hiresense/ingestion/api/routes.py:487,509`
**Confidence:** High

## Description
`get_by_id` runs a synchronous SQLAlchemy session and is called directly in these async handlers, blocking the event loop for the query duration. The rest of the module wraps sync repo work in `asyncio.to_thread`; these two routes are an inconsistency that blocks the loop under load / slow DB.

## Proposed solution
Wrap the sync repo lookups in `asyncio.to_thread`, matching the list endpoint.

## Acceptance criteria
- [ ] Both routes offload sync DB work to a thread

**Estimated effort:** XS

### LINKED ISSUE #158: Performance: getonboard company-name resolution is a sequential N-round-trip loop
**Severity:** Low
**Location:** `backend/src/hiresense/ingestion/adapters/getonboard.py:70-103`
**Confidence:** High

## Description
Each distinct company id is resolved with a separate awaited GET inside a `for raw in jobs` loop with no concurrency. With up to 10 pages × 100 jobs, ingestion serializes hundreds of round-trips; caching dedupes repeats but the first hit of each company is fully serial.

## Proposed solution
Collect distinct company ids and resolve them with a bounded `asyncio.gather` (Semaphore), matching the concurrency pattern used elsewhere.

## Acceptance criteria
- [ ] Company resolution is concurrent with a bounded semaphore

**Estimated effort:** S

### LINKED ISSUE #162: Performance: Cap input and avoid full BeautifulSoup parse per field in strip_html
**Severity:** Low
**Location:** `backend/src/hiresense/ingestion/domain/html_stripper.py:6-15`; callers in normalizers and `hn_hiring.py:34`
**Confidence:** Medium

## Description
Each description runs through a full `BeautifulSoup(html, "html.parser")` parse plus `find_all` and per-tag `insert_before`, once per field per job during bulk ingestion of thousands of jobs, with no size bound. HN's `_is_seeker_post` parses the whole comment just to read the first header line. CPU-heavy on the sync normalize path and unbounded on adversarial HTML.

## Proposed solution
Cap input length before parsing; for the HN header check, avoid a full parse; consider a lighter HTML-to-text pass for the common case.

## Acceptance criteria
- [ ] Input length capped before parsing
- [ ] HN header check avoids full-document parse

**Estimated effort:** S

StevSant:
SSRF hardening plus a batch of ingestion performance fixes, one commit per issue.

### #134 — SSRF in the job-revalidation URL probe (Security, P2)
The revalidation probe (`JobRevalidationService._probe`) fetched `probe_url` — derived from ingested, attacker-influenceable data (boards / HN / CSV) — with `follow_redirects=True`, no scheme/IP validation, and read `resp.text` unbounded.
- New pure `is_safe_probe_url()` (`ingestion/domain/ssrf_guard.py`): http/https only; every resolved A/AAAA record must be globally routable (`not ip.is_global` blocks private, loopback, link-local **incl. 169.254.169.254 metadata**, CGNAT, reserved, multicast; IPv4-mapped IPv6 unwrapped). Fails closed on DNS/parse errors. Resolver is injectable so unit tests run offline.
- `_probe` now follows redirects **manually** (client redirects disabled) and re-validates **every hop**, so an allowlisted host can't bounce the probe to an internal target. Body is streamed and read at most `max_probe_bytes`.
- Limits are configurable: `JOB_REVALIDATION_MAX_PROBE_BYTES` (256 KiB) and `JOB_REVALIDATION_MAX_REDIRECTS` (5). A blocked/over-redirected probe → `UNKNOWN`, never a closure. External-cron contract (`POST /ingestion/revalidate`) unchanged.

### #132 — `/ingestion/jobs` rewrote all scores on every GET (Performance, P2)
`list_jobs` called `persist_scores_batch` with a `ScoreUpdate` for the **entire corpus** on every GET — a full-table UPDATE even on plain pagination, multiplied by the several concurrent list calls the UI fires.
- New pure `changed_score_updates()` (`ingestion/domain/score_change_filter.py`) diffs each job's recomputed `(match_score, semantic_score)` against the value loaded from the store; only genuinely-changed rows are persisted. Write volume is now proportional to real score changes, not corpus size.

### #157 — Sync DB reads block the event loop (Performance, P3)
`analyze_job` and `get_job` called `get_job_by_id` (sync SQLAlchemy) directly in async handlers. Both lookups now go through `asyncio.to_thread`, matching the list endpoint.

### #158 — getonboard company resolution was a serial N-round-trip loop (Performance, P3)
`_resolve_company_names` awaited one `/companies/{id}` GET per job in sequence. It now collects the **distinct** company ids and resolves them concurrently under an `asyncio.Semaphore` (`GETONBOARD_COMPANY_CONCURRENCY`, default 8). Dedup/caching preserved (one round-trip per distinct company).

### #162 — Unbounded full-`BeautifulSoup` parse per field (Performance, P3)
`strip_html` now truncates input to `MAX_HTML_CHARS` (200k) before parsing, bounding CPU/memory on adversarial/huge documents for all ~18 normalizers centrally. HN's `_is_seeker_post` now parses only a bounded header prefix instead of the whole comment.

## Decisions taken
- **#132 — SQL-bounded result set deliberately not done.** The acceptance criterion "result set is bounded (SQL LIMIT/offset)" conflicts with the global ANN pre-rank added in #18 (after this issue was filed), which intentionally materializes the whole corpus to let a high-semantic / low-keyword job reach page 1 before pagination — a SQL LIMIT would break that ranking. The write-amplification (the issue's title, "rewrites all scores on every GET", and the O(N)-per-GET write cost) is fully fixed and covered by a test; the read-side materialization is left as the intended design.
- **#162 — cap as module constants, not new config.** The cap lives as `MAX_HTML_CHARS` / `_HEADER_SCAN_CHARS` module constants (mirroring the existing `MAX_PAGES` domain-constant precedent in the same layer) rather than a settings value threaded through the ~18 pure normalizers, which have no DI/settings seam. `strip_html` still accepts a `max_chars` override.
- **#158 — default concurrency 8.** getonbrd is a normal public API (unlike LinkedIn's aggressive guest endpoint, wired at 1); overridable via env.
- **#134 — `not ip.is_global` as the single block rule** (covers private/loopback/link-local/metadata/CGNAT/reserved at once); manual redirect re-validation; guard injectable to keep unit tests offline.

## Verification
- `uv run ruff check .` → All checks passed
- `uv run ruff format --check .` → 917 files already formatted
- `uv run python -m pytest` (full suite, DB-free) → green. New/changed tests: `test_ssrf_guard.py` (11), `test_job_revalidation_service.py` (+5 SSRF/redirect/body-cap), `test_score_change_filter.py` (4), `test_getonboard_adapter.py` (+1 bounded-concurrency), `test_html_stripper.py` (+2 cap), `test_hn_hiring.py` (+1 long-comment).

## Repo-trap checks
- **Config in groups + .env.example:** `job_revalidation_max_probe_bytes`, `job_revalidation_max_redirects` (ingestion group) and `getonboard_company_concurrency` (job_sources group) added, each with a commented `.env.example` entry.
- **Package `__init__` re-exports:** `is_safe_probe_url` and `changed_score_updates` re-exported from `hiresense.ingestion.domain`.
- **ORM registry / migration:** no new ORM classes and no schema change → none needed.

> Note: `.env.example` and the two config-group files touched here may need trivial conflict resolution at merge against sibling PRs from this batch.

Closes #134
Closes #132
Closes #157
Closes #158
Closes #162


StevSant:
## Self-review (critical pass over the full diff)

Reviewed for correctness, security, and repo conventions. No blocking issues found; a few deliberate trade-offs and residual limitations worth flagging for the human reviewer.

### #134 — SSRF guard (`ssrf_guard.py`, `job_revalidation_service.py`)
- **Residual DNS-rebinding TOCTOU (accepted).** `is_safe_probe_url` resolves the host, then httpx resolves it again at connect time — a rebinding attacker could return a public IP to the guard and a private IP to the client. Pinning the validated address needs a custom transport; documented in the guard docstring. This closes the static-target and cloud-metadata cases (the issue's actual threat), which is the right scope for a P2.
- **Memory bound is ~one transport chunk + `max_probe_bytes`.** `_read_capped` breaks once `total >= max_probe_bytes`, but the final `aiter_bytes` chunk is appended before the check, so peak is `max_probe_bytes` plus one httpx read (~≤64 KiB). Bounded and fine; not literally a hard `max_probe_bytes` ceiling.
- **Retry bypass on probes (intentional).** `.stream(...)` is delegated by `RetryingAsyncClient.__getattr__` to raw httpx, so probes skip retry/backoff. Correct here — a transient error must map to `UNKNOWN`, which is the safe default that never closes a job.
- **`finally` sleeps even on a pre-request block.** Harmless politeness delay when nothing was sent; kept for simplicity/symmetry with the original.
- Redirects: `follow_redirects=False` + manual `urljoin` per hop, each hop re-validated (`job_revalidation_service.py` `_fetch_capped`). Relative and absolute Locations both handled. `_REDIRECT_STATUS` = {301,302,303,307,308}.

### #132 — changed-score diff (`score_change_filter.py`, `routes.py`)
- `original_scores` is snapshotted right after the corpus load and keyed by id, so the later `model_copy` reassignments of `all_jobs` don't affect it. Only the O(N) corpus persist is diffed; the page-level persist after semantic scoring (`routes.py`) is left unconditional — intentionally, it's page-bounded and those rows did change.
- Float compare uses a 1e-9 epsilon; None↔value transitions count as changes. If the pre-ranker ever returns non-deterministic micro-diffs, those rows would still be written — acceptable and self-limiting.
- **Scope note:** the read-side full-corpus materialization is unchanged by design (see PR "Decisions taken" — conflicts with the #18 global pre-rank). So acceptance criterion 2 (SQL-bounded result set) is intentionally not met; criteria 1 and 3 are.

### #157 — `to_thread` wrap (`routes.py`)
Both `analyze_job` and `get_job` wrap `orchestrator.get_job_by_id(...) or scanner.get_job_by_id(...)` in a single `to_thread` lambda; short-circuit semantics preserved.

### #158 — concurrent company resolution (`getonboard.py`)
Distinct ids collected first (dedup preserved → the caching test still sees one `/companies` call per company), resolved under `asyncio.Semaphore`. `_company_name` swallows its own errors (returns ""), so `gather` can't raise and a single failed company degrades to blank, matching prior behaviour.

### #162 — input cap (`html_stripper.py`, `hn_hiring.py`)
- Cap applied centrally in `strip_html` → covers all ~18 normalizers. Truncation can cut mid-tag, but `html.parser` tolerates partial markup.
- HN `_is_seeker_post` parses only `text[:_HEADER_SCAN_CHARS]`; the seeker/company decision reads the first header line, which is always at the start, so behaviour is preserved (existing seeker/company tests + a new long-comment test cover it).
- **Decision:** cap is a module constant (mirrors the existing `MAX_PAGES`), not threaded config — the normalizers have no settings seam. See PR body.

### Conventions
Package re-exports added (`is_safe_probe_url`, `changed_score_updates`); new settings live in the matching config groups + `.env.example` with comments; no new ORM class or schema change → no migration/registry edit. `ruff check` + `ruff format --check` clean repo-wide.

StevSant (commented):
## Code review — #199

**Done well**
- `backend/src/hiresense/ingestion/domain/ssrf_guard.py` — `is_safe_probe_url` uses `not ip.is_global` (via `_is_globally_routable`) rather than a private-only check, so it correctly blocks loopback, link-local (incl. `169.254.169.254` cloud metadata), CGNAT (`100.64/10`), reserved and multicast in one rule. It unwraps IPv4-mapped IPv6 (`::ffff:10.0.0.1`) before classifying, requires *every* resolved A/AAAA record to be public (blocks a mixed public+private DNS answer), and fails closed on `OSError`/`ValueError`/`UnicodeError` from the resolver. Resolver is injectable, keeping `test_ssrf_guard.py` fully offline (11 cases, including the metadata endpoint, CGNAT, mixed-answer, and IPv4-mapped cases).
- `job_revalidation_service.py`'s `_fetch_capped` follows redirects manually with `follow_redirects=False` and re-validates **every** hop (including the first) before connecting — confirmed via `test_probe_blocks_redirect_to_private_target`, which shows a public first hop is fetched but a 302 to `10.0.0.5` is refused before the internal URL is ever requested. Body reads are capped via streaming (`test_probe_body_read_is_capped`), and a blocked/over-redirected probe correctly returns `UNKNOWN` (never a closure) via `_ProbeBlocked`.
- `#132` fix (`score_change_filter.py` + `routes.py`): `changed_score_updates` diffs against a snapshot (`original_scores`) taken right after the corpus load, before skill-scoring/pre-rank mutate `all_jobs`, so the comparison is against the actually-persisted value. `None`↔value transitions correctly count as changes; near-equal floats within `1e-9` don't. Good test coverage including a "steady-state emits nothing" case.
- `#157`: `analyze_job`/`get_job` correctly wrap the *sync* `orchestrator.get_job_by_id(...) or scanner.get_job_by_id(...)` call in a `lambda` passed to `asyncio.to_thread` — this offloads a blocking call, not an already-async one.
- `#158`: `_resolve_company_names` dedupes to distinct company ids first, then resolves them concurrently under a `Semaphore(company_concurrency)` — one round-trip per distinct id, bounded by `GETONBOARD_COMPANY_CONCURRENCY`. `test_getonboard_resolves_companies_concurrently_within_bound` verifies actual overlap (`max_inflight == 2` with a bound of 2, not 1) rather than just call count.
- `#162`: `strip_html`'s truncate-before-parse is applied centrally, and `hn_hiring._is_seeker_post` correctly still classifies long seeker posts since the marker is at the very start of the comment (`test_long_seeker_comment_still_rejected`).
- New settings (`job_revalidation_max_probe_bytes`, `job_revalidation_max_redirects`, `getonboard_company_concurrency`) are properly threaded through `config/groups/*.py` → `bootstrap/ingestion.py` → `.env.example` with explanatory comments, and new public symbols (`is_safe_probe_url`, `changed_score_updates`) are re-exported from `ingestion/domain/__init__.py`, per repo convention.

- 🟠 `backend/src/hiresense/ingestion/domain/job_revalidation_service.py:222` — Redirect handling switched from `self._http.get(probe_url, follow_redirects=True, ...)` to `self._http.stream("GET", current, follow_redirects=False, ...)`. The shared HTTP client is `RetryingAsyncClient`, which only wraps `get`/`post`/`request` with retry+backoff on transport errors and on `[429, 500, 502, 503, 504]` (`backend/src/hiresense/adapters/http/retrying_async_client.py:55-84`); `stream` is not overridden and is delegated straight through `__getattr__` to the raw `httpx.AsyncClient`, bypassing retry entirely. Previously a transient 503 or connection reset on a probe would be retried (up to `http_max_retries`, default 3, with backoff) before falling back to UNKNOWN; now it goes to UNKNOWN on the very first transient failure. This is fail-safe (never causes a false closure) so it's not a correctness bug, but it silently reduces the sweep's ability to actually detect real closures on flaky hosts — the same class of problem the existing custom User-Agent already works around for weworkremotely-style hosts — and isn't mentioned in the PR description's verification section. Worth a one-line callout or a follow-up issue if intentional; if unintentional, consider adding a small retry loop around `_fetch_capped`'s per-hop stream call, or a retry-aware `stream()` on `RetryingAsyncClient`.

**Verified**
- Read `ssrf_guard.py`, `job_revalidation_service.py`, `score_change_filter.py`, `routes.py`, `getonboard.py`, `hn_hiring.py`, `html_stripper.py`, `bootstrap/ingestion.py`, both touched `config/groups/*.py` files, and `.env.example` in full (via `git show origin/fix/ingestion-perf-ssrf-batch:<path>`).
- Confirmed `RetryingAsyncClient.stream` is not overridden (delegates via `__getattr__`) and that the pre-PR code path used `.get()` (which *is* wrapped), establishing the retry-loss finding above.
- Confirmed `httpx.Response.encoding` (used in `_read_capped`) does not require the body to be fully buffered when `default_encoding` is the default string `"utf-8"` — no `ResponseNotRead` risk from accessing `.encoding` on a partially-streamed response.
- Traced the redirect-cap loop (`range(self._max_redirects + 1)`): with `max_redirects=5` it makes at most 6 requests, following at most 5 redirects before raising `_ProbeBlocked`, matching the documented semantics and the loop/cap tests (`test_probe_redirect_loop_exceeding_cap_is_unknown`).
- Verified all new config fields are wired through `config/groups/ingestion.py` and `config/groups/job_sources.py` into `bootstrap/ingestion.py` and documented in `.env.example` (no hardcoded values).
- Verified `changed_score_updates` is only applied at the corpus-wide persist site (`routes.py`, the `#132` fix); the separate page-level `persist_scores_batch` call after page-level semantic scoring is unchanged and out of scope (already bounded to page size).
- Reviewed all new/changed unit tests (`test_ssrf_guard.py`, `test_job_revalidation_service.py`, `test_score_change_filter.py`, `test_getonboard_adapter.py`, `test_hn_hiring.py`, `test_html_stripper.py`) — coverage matches the claims in the PR description (SSRF block/allow/redirect/loop/body-cap, score-diff steady-state, bounded concurrency with observed overlap, long-comment truncation, input-cap truncation).

Review by Claude Code

PR #200, opened 2026-07-20
perf(applications): aggregate query batching, list pagination, lazy skill fallback

### LINKED ISSUE #144: Performance: N+1 and duplicated queries building application aggregates
**Severity:** Medium
**Location:** `backend/src/hiresense/applications/domain/application_service.py:121-228`; `api/routes.py:82-126`
**Confidence:** High

## Description
`_build_aggregate` issues ~9 queries per application: `get_latest_match` calls `list_matches` (loads all rows, takes [0]) then `match_count=len(list_matches(...))` runs the same query again; the double-query pattern repeats for optimizations, interview_preps, cover_letters, plus a snapshot lookup. `list()` runs `_build_aggregate` per application, and `_to_list_item` calls `orchestrator.get_job_by_id` per row. The list endpoint is O(N × ~10) queries, unbounded.

## Proposed solution
Add repository methods returning latest-row + count in one query each (`ORDER BY created_at DESC LIMIT 1`, `count_*`), or batch-load children for the requested ids; use a lightweight projection for the list view and batch the job enrichment.

## Acceptance criteria
- [ ] Per-application query count is constant, not ~10
- [ ] List endpoint batches job enrichment
- [ ] Query-count assertion in a test

**Estimated effort:** M

### LINKED ISSUE #145: Performance: Add pagination / bounds to application, cover-letter, and tracking list endpoints
**Severity:** Medium
**Location:** `applications/api/routes.py:48-53,82-88`; `tracking/api/routes.py:70-77`; `applications/infrastructure/repository.py:182-215`
**Confidence:** High

## Description
`GET /applications`, `GET /applications/cover-letters`, and `GET /tracking` return every row with no LIMIT/offset. `list_all_cover_letters_with_context` returns full cover-letter bodies for every application in one unbounded query. Combined with the aggregate N+1 above, response size and query count grow without bound as history accumulates.

## Proposed solution
Add `limit`/`offset` (or cursor) params with sane defaults and a SQL LIMIT; return counts in metadata per the API-envelope convention.

## Acceptance criteria
- [ ] All three endpoints bounded and paginated
- [ ] Pagination metadata returned
- [ ] Tests cover paging

**Estimated effort:** M

### LINKED ISSUE #171: Performance: Compute the SkillMatcher fallback lazily in generate_match
**Severity:** Low
**Location:** `backend/src/hiresense/applications/domain/artifact_service.py:67-73`
**Confidence:** High

## Description
`generate_match` always constructs a `SkillMatcher` and runs `.match(...)` (normalization + set math) even when `result.matched_skills`/`missing_skills` are already populated — the fallback is only used when they're empty. Wasted work on every match generation.

## Proposed solution
Compute the fallback lazily only when `matched_skills`/`missing_skills` are empty.

## Acceptance criteria
- [ ] Matcher runs only when the fallback is needed

**Estimated effort:** XS

StevSant:
## Summary

Performance batch for the `applications` (and `tracking`) modules — three related issues that all made the application-list surface slower and heavier as history grows.

### #144 — N+1 and duplicated queries building application aggregates
`_build_aggregate` issued ~10 queries per application: every `get_latest_*` loaded **all** child rows and took `[0]`, then each `*_count` re-ran the same full query, and `_to_list_item` resolved the linked job one row at a time. On the list endpoint that was O(N × 10) unbounded queries.

- `get_latest_match/optimization/interview_prep/cover_letter` now use `ORDER BY created_at DESC LIMIT 1` instead of load-all-then-`[0]`.
- New batch child loaders (`get_snapshots_for` / `list_*_for`) group rows by `application_id` in **one query each**; `ApplicationService` builds every aggregate from these dicts, so the whole list costs a fixed **5 queries regardless of N** (single `get` also drops from ~10 to 5).
- New `IngestionOrchestrator.get_jobs_by_ids` (+ `JobsRepositoryPort.get_by_ids` on both repositories) resolves job enrichment once per list request instead of per row.
- Regression test asserts the list stays at 5 queries for N=1 and N=4.

### #145 — Pagination / bounds on application, cover-letter, and tracking list endpoints
`GET /applications`, `GET /applications/cover-letters`, and `GET /tracking` returned every row with no LIMIT (cover letters returned full bodies for every application).

- Optional `limit`/`offset` on the repository/service list methods (default `None` = unbounded) with DB-side `LIMIT`/`OFFSET`.
- Deterministic ordering (`created_at DESC, id`) so pages don't overlap — `created_at` ties at second precision on SQLite.
- Routes read `?limit`/`?offset` (generous config default, clamped to a max) and return the total in an **`X-Total-Count`** response header, exposed via CORS.
- New config `default_page_size` / `max_page_size` (core group + `.env.example`); shared `kernel.resolve_page_limit` helper.

### #171 — Compute the SkillMatcher fallback lazily in generate_match
`generate_match` always built a `SkillMatcher` and ran `.match()` even when `analyze()` already returned both `matched_skills` and `missing_skills`. The fallback now runs only when a verdict field is empty; behaviour is unchanged (still per-field preference of the analyze verdict).

Closes #144
Closes #145
Closes #171

## Decisions taken

- **#145 metadata via header, not envelope.** The issue says "return counts in metadata per the API-envelope convention", but all three endpoints already return bare arrays that the Angular frontend (`applications.service.ts`, `tracking.service.ts`) consumes directly. Switching to a `{data, meta}` envelope would be a breaking change across the frontend. Instead the total travels in an `X-Total-Count` header (standard REST convention, exposed via CORS), keeping the response shape 100% backward-compatible — no frontend changes required. A future frontend paginator can read the header.
- **Generous default page size (100), applied at the route layer only.** Internal callers (`analytics` funnel/benchmark, `bootstrap/inbox`, `matching` batch-evaluate) call the service `.list()` expecting the full set, so the service/repo default stays `None` (unbounded); only the HTTP routes impose the default bound. For a single-user job-hunt tool a 100-row default page is generous; realistic datasets sit well under it, and the true total is always in `X-Total-Count`.
- **`get_latest_*` kept as public repo methods** (still used by artifact/apply services) and optimized to `LIMIT 1`, rather than removed — the batch builder no longer calls them but other call sites do.

## Verification

- `uv run python -m pytest` — full backend suite green (DB-free SQLite path).
- `uv run ruff check .` — clean on touched dirs.
- `uv run ruff format --check` — clean (3 files formatted during work).
- Query-count regression test proves the list is a constant 5 queries for N=1 and N=4.
- Frontend untouched: bare-array response shapes preserved, so no Angular consumer changes and no frontend build/test needed.

## Repo-trap checks

Package `__init__` re-exports updated (`kernel.resolve_page_limit`); new config in the matching `config/groups/core.py` **and** `.env.example`; no new ORM classes (no `registry.py`/migration needed); no hardcoded limits (page sizes are config); domain layer imports no framework/infrastructure.

## Note

Sibling batch PRs may create trivial conflicts in `.env.example` / `config/groups/core.py` (shared config files).


StevSant:
### Self-review (critical)

**Correctness**
- `_build_aggregates` batch loaders order child rows by `created_at DESC` only when picking `[0]` as "latest" (`repository.py` `_children_for`). Two rows of the same application sharing a second-precision `created_at` make "latest" ambiguous — but this matches the pre-existing `get_latest_*` behaviour (also `created_at DESC` only), so it's not a regression. The list-level pagination (`list_all` / cover-letter listing) *does* add an `id` tiebreaker where page stability actually matters.
- `count()` / `count_all_*` add one extra COUNT query per list request for `X-Total-Count`. Intentional and bounded (1 query), the standard price of returning a total.
- `if offset:` correctly no-ops on `offset=0` and `offset=None`; `limit=None` means unbounded, preserved for internal callers.

**Performance / trade-offs**
- The unified single+list aggregate builder means the **list** path loads full cover-letter `body` text via `list_cover_letters_for` just to compute latest+count, then drops the body in `_to_list_item` (`routes.py`). Bounded by `max_page_size` (500) so acceptable, but a lightweight count-only projection would avoid loading bodies on the list. Chose the unified builder for simplicity; flagging the tradeoff.
- Tracking's `_enrich` (`tracking/api/routes.py`) still resolves the linked job per-row via `get_job_by_id`. #144 scoped job-enrichment batching to `applications`; the same N+1 pattern remains in tracking but is now bounded by the page size. Reasonable follow-up: reuse `get_jobs_by_ids` there too.

**Conventions**
- New config in `config/groups/core.py` + `.env.example`; no hardcoded page sizes; shared helper re-exported from `kernel`. Domain layer takes no framework imports. No new ORM ⇒ no migration.

**Behaviour change to call out**
- `GET /applications|/tracking|/cover-letters` now return at most `default_page_size` (100) rows when the client sends no `?limit`. The Angular frontend sends none, so a user with >100 rows would see only the first page until the frontend reads `X-Total-Count` and paginates. Deliberate per the "generous defaults" steer; total is always exposed.

StevSant (commented):
## Code review — #200

**Done well**
- The N+1 fix is real and verified by a query-count regression test (`backend/tests/unit/applications/test_list_query_count.py`), which asserts a constant 5 queries for N=1 and N=4 applications — a strong guard against future regressions.
- `_build_aggregates`/`_assemble` correctly reconstruct per-application "latest" and "count" from the batch-loaded, `created_at`-descending grouped dicts, with no extra round-trips (`backend/src/hiresense/applications/domain/application_service.py`).
- The lazy `SkillMatcher` fallback in `generate_match` (`backend/src/hiresense/applications/domain/artifact_service.py:63-73`) is provably behavior-preserving: I traced all four truthy/falsy combinations of `result.matched_skills`/`result.missing_skills` against the old always-compute-fallback code, and the new short-circuit produces identical output in every case. Backed by two new tests that spy on `SkillMatcher.__init__`/`.match` to assert it's skipped/run exactly when expected.
- Pagination ordering ties are broken by `id` (`created_at.desc(), id`) specifically because `created_at`'s server default has second-precision on SQLite — correctly identified and tested (`test_list_all_cover_letters_paginates_and_counts` checks non-overlapping pages across a boundary).
- `resolve_page_limit` / batch `get_by_ids` helpers correctly no-op on empty input, and default `limit=None`/`offset=None` at the service/repo layer keeps every internal caller (`bootstrap/inbox.py`, `matching/api/routes.py`, `analytics/domain/comp_benchmark_service.py`) on the old unbounded behavior since they never pass the new kwargs.
- New `.env.example` / `config/groups/core.py` entries follow the project's config conventions (no hardcoded page sizes), and `CORS expose_headers=["X-Total-Count"]` is correctly added in `main.py` — an easy step to forget when moving data into a response header.
- CI is green end-to-end (format, lint, tests, alembic, pgvector, build, frontend, GitGuardian).

- 🟠 `frontend/src/app/core/services/applications.service.ts:23`, `tracking.service.ts:14-19` — none of the three paginated endpoints (`/applications`, `/applications/cover-letters`, `/tracking`) are called with `?limit`/`?offset` from the frontend, and nothing in `frontend/src/app/**` reads the new `X-Total-Count` header (`grep` for it returns no matches). That means the previously-unbounded lists are now silently capped at `default_page_size` (100) with no UI signal that more rows exist. This is a documented, deliberate tradeoff in the PR body ("generous default... realistic datasets sit well under it"), and 100 is unlikely to bite immediately, but it's a real behavior change (unbounded → truncated) for a tool whose own project history includes an autopilot pipeline that grows the tracked-applications table unattended. Worth a tracked follow-up issue for the frontend to consume the header (or add "load more"/paging UI) rather than leaving it as an implicit assumption.
- 🟡 `backend/src/hiresense/applications/api/routes.py:20-33` vs `backend/src/hiresense/tracking/api/routes.py:20-25,74-79` — the "resolve page size from `request.app.state.settings` with a bare-app-test fallback constant" logic is duplicated near-verbatim in both route modules (including the `_DEFAULT_PAGE_SIZE = 100` / `_MAX_PAGE_SIZE = 500` constants). Applications' version is a named `_page_size(request, limit)` helper; tracking's is inlined in the route body. Low risk since both are tiny and read from the same config fields, but a shared `kernel` helper (e.g. `resolve_page_limit_from_request`) would remove the duplicate constants and keep the two endpoints from drifting if the fallback logic ever needs a fix.
- 🟡 `backend/src/hiresense/tracking/api/routes.py:76` (`_enrich(a, orchestrator)` per row in `list_applications`) — the tracking list endpoint still resolves job enrichment with one `orchestrator.get_job_by_id(...)` call per row (pre-existing code, untouched by this PR), unlike the applications list endpoint which now batches via the new `get_jobs_by_ids`. Out of scope for #144/#145/#171 as written, but since the batch helper now exists it'd be a small, obvious follow-up to close the same N+1 gap here.

**Verified**
- Full diff across all 25 changed files (`git diff origin/main...origin/perf/applications-batch`), including repository, service, route, port, config, and test layers for `applications`, `tracking`, `ingestion`, and `kernel`.
- Manually replayed the `get_latest_*` → `LIMIT 1` conversions and the `_build_aggregates` batch rewiring against the pre-PR code — no other call site of the per-item repo methods lost correctness (`apply_service.py` and `artifact_service.py` still legitimately use the single-item `get_latest_*` methods for one-off generate/apply flows, now benefiting from `LIMIT 1` instead of load-all-then-`[0]`).
- Confirmed `resolve_page_limit`'s `requested is None → default` / `else → min(requested, maximum)` semantics match the route-level `Query(None, ge=1)` validation (limit=0 rejected with 422, offset<0 rejected with 422) via `test_list_rejects_invalid_pagination`.
- Confirmed the SkillMatcher lazy-fallback truth table by hand against the old code (see "Done well").
- Confirmed `CoreSettings.default_page_size`/`max_page_size` match the route fallback constants and `.env.example` values (100/500 in both places).
- `gh pr checks 200` — all green (backend format/lint/tests/alembic/pgvector/build, frontend format/typecheck/a11y/tests/build, GitGuardian).

**Verdict:** LGTM ✅

Review by Claude Code

PR #201, opened 2026-07-20
fix(ingestion): scoring correctness batch — id-keyed quick scores, cooldown race, weight renormalization, embedding-text dedup

### LINKED ISSUE #143: Bug: Quick-scoring positional fallback can attribute an LLM score to the wrong job
**Severity:** Medium
**Location:** `backend/src/hiresense/ingestion/domain/quick_scoring_service.py:200-231`
**Confidence:** Medium

## Description
When the LLM omits/garbles `ref`, `_resolve_job` falls back to `chunk[idx]` — mapping the Nth returned object to the Nth job in the chunk. But `_parse` enumerates over the model's array, which may be shorter/longer/reordered. If the model returns fewer objects or a different order without refs, scores/verdicts/dealbreakers bind to the wrong jobs, then get cached per `(job_id, profile_hash)` and shown as authoritative — e.g. "Requires Java — not in your stack" on a job that never mentioned Java.

## Proposed solution
Trust only range-validated `ref`; drop items without a usable ref instead of positional guessing. If positional fallback is kept, apply it only when the returned count equals the chunk length.

## Acceptance criteria
- [ ] Mismatched/reordered LLM output cannot bind a score to the wrong job
- [ ] Test with reordered/short LLM output asserts correct or dropped mapping

**Estimated effort:** S

### LINKED ISSUE #159: Bug: Ingestion cooldown check is a check-then-set race under concurrent /fetch
**Severity:** Low
**Location:** `backend/src/hiresense/ingestion/domain/services.py:66-71`
**Confidence:** Medium

## Description
The cooldown reads `self._last_run_at` and later assigns it with no lock. Two concurrent `/fetch` calls can both observe the pre-cooldown value and both proceed, doubling outbound fetch load and duplicating the ingestion pass. Also `_last_run_at` is set before the work, so a run that fails immediately still consumes the cooldown window.

## Proposed solution
Guard check-and-set with an `asyncio.Lock` (or in-flight flag); set the timestamp on successful completion, or use a single-flight guard so a second concurrent trigger is a no-op.

## Acceptance criteria
- [ ] Concurrent `/fetch` cannot double-run the pass
- [ ] Cooldown timestamp reflects successful completion

**Estimated effort:** S

### LINKED ISSUE #160: Bug: combine_fit_score ignores its weights when one signal is missing, biasing the sort
**Severity:** Low
**Location:** `backend/src/hiresense/ingestion/domain/job_scorer.py:42-48`
**Confidence:** Medium

## Description
When only one of skill/semantic is present, the function returns that raw signal unweighted. A job with only a skill score (0.8) ranks on 0.8, while a job with both ranks on `0.4*skill + 0.6*semantic` (smaller for the same skill). The global `/jobs` sort systematically advantages single-signal jobs over jobs with both. The pre-ranker only sets semantic on ANN-indexed jobs, so this skews the mix.

## Proposed solution
Define an explicit single-signal contract (treat missing as 0, or renormalize by present weights) and apply consistently; document it. Add a mixed present/absent test.

## Acceptance criteria
- [ ] Consistent, documented blending rule for single-signal jobs
- [ ] Test covers the mixed case

**Estimated effort:** S

### LINKED ISSUE #161: Refactor: Consolidate duplicated embedding-text/cache logic across scoring services
**Severity:** Low
**Location:** `ingestion/domain/semantic_scoring_service.py:18-29,85-101`; `semantic_pre_ranker.py:39-41,186-207`; `job_embedding_indexer.py:13-15`
**Confidence:** High

## Description
`_job_text` (title+skills+description, 4000-char cap) is copy-pasted in `semantic_scoring_service` and `job_embedding_indexer`; `_profile_key` (sha256 of "skills\nsummary") and the "embed profile once, cache by hash, degrade on empty/failure" flow are near-identical between `SemanticScoringService` and `SemanticPreRanker`. If index-time job text drifts from query-time text, ANN and in-memory cosine scores silently disagree — a correctness footgun, not just DRY.

## Proposed solution
Extract a shared helper for `job_text`, `profile_text`, `profile_key`, and a cached-profile-embedding helper; have all three consumers import it.

## Acceptance criteria
- [ ] Single source for job/profile embedding text and profile key
- [ ] Index-time and query-time job text guaranteed identical

**Estimated effort:** S

StevSant:
Scoring-correctness batch for the ingestion bounded context: four issues, one commit each, all pure/domain-layer changes with unit coverage.

### #143 — Quick-scoring positional fallback could bind an LLM score to the wrong job
`QuickScoringService._parse` bound the Nth returned object to the Nth job whenever the model omitted a `ref`, so a shorter/reordered array silently attributed a score/verdict/dealbreakers to the wrong job (then cached as authoritative). Fix: trust only a range-validated `ref`; keep the positional fallback **only** when the returned array is 1:1 with the chunk (`len(data) == len(chunk)`), otherwise drop the ambiguous item. Items that carry a valid `ref` still map correctly even when the count differs.
- Tests: reordered-with-refs maps by ref; short ref-less output is dropped (not guessed); short output with valid refs still maps by ref; equal-count ref-less output still maps positionally.

### #159 — Ingestion cooldown was a check-then-set race under concurrent `/fetch`
`run()` read and later wrote `self._last_run_at` with no guard, so two concurrent triggers could both clear the cooldown and double-run the pass; the timestamp was also set *before* the work, so a fast failure still consumed the window. Fix: a single-flight boolean guard claimed atomically (no `await` between the in-flight check, the cooldown check, and the claim, so it is atomic on the event loop), the timestamp is now set only after a fully successful pass, and the guard is released in `finally` **only if this call claimed it** (a cooldown-rejected concurrent trigger never clears the in-flight run's flag).
- Tests: concurrent runs execute the pass exactly once; a failed run neither starts the cooldown nor sticks the guard (retry reaches the work again); a successful run starts the cooldown.

### #160 — `combine_fit_score` ignored its weights when one signal was missing
A single-signal job was returned raw while a two-signal job used the weighted blend, giving single-signal jobs (the non-ANN-indexed ones) an inconsistent, undocumented advantage in the global sort. Fix: an explicit, documented single-signal contract — the blend renormalizes over the *present* weights (the lone signal's weight scales to 1.0), keeping every job on the same [0, 1] scale. Two-signal behaviour with default weights (summing to 1.0) is unchanged. Returns `None` when no signal is present or the present weights sum to zero.
- Tests: new `test_combine_fit_score_renormalization.py` covers skill-only, semantic-only, the mixed present/absent case on a shared scale, the two-signal no-op, and the zero-weight → None edge.

### #161 — Duplicated embedding-text / profile-cache logic across scoring services
`_job_text` was copy-pasted in `semantic_scoring_service` and `job_embedding_indexer`, and `_profile_key` / the "embed once, cache by hash, degrade on empty/failure" flow were near-duplicated in `SemanticScoringService` and `SemanticPreRanker` — and had already **drifted**: the pre-ranker hashed and embedded the profile text **without** the 4000-char cap the scoring service applied, so the same profile could yield different keys/vectors and disagreeing scores. Fix: a new `ingestion/domain/embedding_text/` package (one function per file, re-exported from the package `__init__`) with `job_text`, `profile_text`, `profile_key`, and `embed_profile_cached`; all three consumers now import from it. Index-time and query-time job text are now guaranteed identical, and both profile embeddings use the same capped text. Each service keeps its own lock semantics at the call site.

Closes #143
Closes #159
Closes #160
Closes #161

## Decisions taken
- **#160 blend rule:** chose *renormalize over present weights* (not *treat missing signal as 0*). Treating missing-as-0 would systematically bury every non-ANN-indexed job (which never has a semantic signal) — a large, undesirable ranking shift — whereas renormalization keeps jobs comparable on one scale. This matches the issue's recommended option and the pre-ranker's existing effective behaviour, now made explicit and documented.
- **#143 fallback:** kept the positional fallback but gated it on an exact count match, per the issue's "if positional fallback is kept, apply it only when the returned count equals the chunk length." Fully dropping the fallback would have regressed the common ref-less-but-in-order response.
- **#161 char limits:** kept the existing `4000` caps as named module constants co-located with the shared functions (single source), rather than introducing new config — they are internal truncation caps, not operator-tunable thresholds, and this matches the pre-existing pattern.
- **#159 guard:** used an in-flight boolean (sanctioned by the issue) rather than an `asyncio.Lock`; on the single-threaded event loop the claim is atomic and a second trigger is rejected rather than queued.

## Verification
- `uv run ruff check .` → All checks passed.
- `uv run ruff format --check` on all touched files → 14 files already formatted.
- Targeted ingestion unit tests (quick-scoring, combine_fit_score ×3, orchestrator, semantic services, indexer) → 75 passed.
- Full backend suite → **1381 passed, 6 skipped, 1 failed**. The single failure is `test_settings_autopilot::test_autopilot_pipeline_settings_defaults`, a known-environmental failure from the local `.env` setting `AUTOPILOT_PIPELINE_ENABLED=true`; it is unrelated to this change and passes in CI (no local `.env`).

## Repo-trap checks
Used `uv run python -m pytest` (not bare `uv run pytest`); switched to StevSant before every `gh`; did **not** run `ruff format .` repo-wide (only checked/formatted touched files, since the repo is not format-clean and CI runs only `ruff check`); imports go through the contextual package (`from hiresense.ingestion.domain.embedding_text import …`); one function per file; no new hardcoded config values.

## Overlap with sibling PR #199 (`fix/ingestion-perf-ssrf-batch`)
Low conflict risk. #199 touched `list_jobs` score persistence (`changed_score_updates`), `strip_html`, the revalidation service, and the getonboard adapter — none of which this PR modifies. My `services.py` change is confined to the `run()` cooldown gate and does not touch `list_jobs`/`persist_scores`. `combine_fit_score` (used by the pre-ranker sort path) is the only shared conceptual surface; the change is additive and behaviour-preserving for the two-signal case, so no functional conflict is expected even if the files are edited on both branches.


StevSant:
## Self-review (critical)

Reviewed for correctness, hexagonal conventions, and behaviour changes. No blockers found; flagging the surfaces a reviewer should scrutinize.

**#143 — `quick_scoring_service.py` `_parse`/`_resolve_job`**
- The count guard `positional_ok = len(data) == len(chunk)` counts non-dict array items toward the length. That is still correct: `enumerate(data)` keeps each dict's index aligned with its `chunk` slot, and a skipped non-dict simply yields no result at its position. Ref-carrying items always map by ref regardless of the guard.
- Known, accepted limitation: a **reordered-but-ref-less** response whose length happens to match the chunk still maps positionally — unresolvable without refs, and the system prompt explicitly requires `ref`, so this is the pragmatic floor the issue sanctioned, not a correctness hole.

**#159 — `services.py` `run()`**
- The single-flight boolean is atomic only under single-threaded event-loop semantics (async FastAPI handlers) — correct for how `run()` is invoked. It does not serialize across threads/processes, but neither did the previous per-process `self._last_run_at`; cross-process throttling was never offered, so this is not a regression. Worth a reviewer's eye if a multi-worker deployment ever needs global throttling.
- `finally` releases the guard only when `claimed` is True, so a cooldown-rejected concurrent trigger can't clear the in-flight run's flag. The timestamp now advances only on a fully successful pass.

**#160 — `job_scorer.py` `combine_fit_score`**
- Behaviour change to flag: a single present signal whose weight is **0** now returns `None` (previously returned the raw score). No production caller passes a zero weight for a single-signal case — the pre-ranker always has the semantic signal present with positive `prerank_weight_*`. Two-signal results with default weights are bit-identical to before; the single-signal path returns the raw score exactly (special-cased to avoid `w*s/w` float drift, which was caught by `test_only_skill_returns_skill`'s exact `== 0.7`).

**#161 — `embedding_text/` package + three consumers**
- Intended behaviour change: `SemanticPreRanker` now caps profile text at 4000 chars (previously uncapped) and its profile cache key changes for profiles >4000 chars. This is the fix — it aligns the pre-ranker's key/vector with `SemanticScoringService` so ANN and in-memory cosine scores stop disagreeing. The cache is in-process and ephemeral, so no migration.
- `embed_profile_cached` adds an empty-vector guard (`if not vec: return None`) to the scoring-service path that previously lacked it — strictly safer.
- Lock semantics preserved at each call site: `SemanticScoringService` holds `self._lock` across the call in `score_jobs`; `SemanticPreRanker._get_profile_embedding` wraps the call in `async with self._lock`. The shared helper is intentionally lock-free (documented).
- Convention check: package `__init__` re-exports all four functions; consumers import from `hiresense.ingestion.domain.embedding_text` (contextual package), not the implementation files; one function per file. Domain purity intact — no `infrastructure`/framework imports.

StevSant (commented):
## Code review — #201

**Done well**
- `#143` fix (`quick_scoring_service.py`): gating the positional fallback on `positional_ok = len(data) == len(chunk)` is exactly right — hand-traced `_resolve_job`: when counts differ, only items with a range-validated `ref` map to a job; ref-less items are dropped instead of guessed. All four new tests (`test_reordered_output_maps_by_ref_not_position`, `test_short_refless_output_is_dropped_not_positionally_guessed`, `test_short_output_with_valid_refs_still_maps_by_ref`, `test_equal_count_refless_output_still_maps_positionally`) match the implementation's actual branching.
- `#159` fix (`services.py`): the in-flight check, cooldown check, and claim (`self._run_in_flight = True; claimed = True`) all run with no `await` between them, so on the single-threaded event loop they're genuinely atomic — a concurrent `run()` cannot slip through. `_last_run_at` is now set only on the success path, and `finally` releases the guard only `if claimed`, so a cooldown-rejected caller never clears the in-flight run's flag. Verified the statement order: `_last_run_at` is set *before* `finally` clears `_run_in_flight`, so there's no gap where both checks are simultaneously bypassable. `test_concurrent_runs_do_not_double_run` uses a genuinely gated fake source (`asyncio.Event`) rather than mocking away the concurrency, which is the right way to test this.
- `#161` dedup: confirmed against the pre-PR code that this was a real bug, not just a style issue — `semantic_pre_ranker._profile_key` hashed the **uncapped** `skills+summary` text while `semantic_scoring_service._profile_key` hashed the **capped** (4000-char) version, so the same profile could produce different cache keys/vectors between the two services. The new `embedding_text/profile_key.py` derives the key from the same `profile_text()` that gets embedded, closing that gap. `job_text` extraction from `job_embedding_indexer`/`semantic_scoring_service` is byte-identical to both originals — a clean, behavior-preserving dedup. The new package is one function per file with a re-exporting `__init__.py`, matching repo convention.
- `#160`: hand-verified the arithmetic. For the single-signal case, `weighted[0]` reduces to `score if weight > 0 else None`, which is mathematically identical to the old "return the raw signal" fallback for any positive weight — so this is correctly framed as formalizing/documenting an existing (and, per the linked issue, intentionally-kept) behavior rather than changing rankings, and matches the issue's own "Low" severity/effort estimate. New `test_combine_fit_score_renormalization.py` covers all the documented cases and the assertions match hand-computed values (e.g. `combine_fit_score(1.0, 0.0) == 0.4` with default weights).
- No regression risk to the min_score/getonboard interaction noted in prior audits: `job_filter.py` isn't touched by this PR, and `combine_fit_score`'s single-signal-missing return value is numerically unchanged, so that interaction is neither improved nor worsened here.

- 🟠 `backend/src/hiresense/ingestion/domain/embedding_backfill_service.py:14-19,77` — this file has its own separate, byte-identical copy of the job-text formula that #161 set out to eliminate, and this PR doesn't touch it:
  ```python
  _JOB_TEXT_CHAR_LIMIT = 4000

  def _job_text(job: NormalizedJob) -> str:
      parts = [job.title, " ".join(job.skills), job.description]
      return "\n".join(p for p in parts if p)[:_JOB_TEXT_CHAR_LIMIT]
  ```
  This is the backfill-embedding path (`POST /ingestion/backfill-embeddings`), which is also an index-time producer of job vectors in the same vector store `SemanticScoringService`/`JobEmbeddingIndexer` query against. It's in sync today only because nobody has touched either copy since the fork. A future edit to the job-text formula in the new `ingestion/domain/embedding_text/job_text.py` will silently miss this file, reintroducing exactly the index/query drift risk #161 just fixed for the other two services. Recommend importing `job_text` from `hiresense.ingestion.domain.embedding_text` here too and dropping the local copy.

- 🟡 `backend/src/hiresense/ingestion/domain/job_scorer.py:44-51` — the renormalization now applies to the two-signal branch as well, not just the single-signal one: `sum(weight*score) / total_weight`. With the default weights (0.4/0.6, summing to 1.0) this is a no-op vs. the old `w1*s1 + w2*s2`, but `prerank_weight_skill`/`prerank_weight_semantic` (`config/groups/matching.py:12-13`) have no validator enforcing they sum to 1.0 — unlike the sibling 10-dimension `weight_*` group, which is explicitly documented as "must sum to 100." If an operator sets non-normalized values, the two-signal blend result changes from pre-PR behavior (now scaled up to compensate for weights below 1.0, or down for weights above 1.0). The docstring's "existing behaviour is unchanged" claim is only true for weights summing to 1.0 — worth either adding a sum-to-1 validator on those two settings or calling out in the docstring that non-normalized two-signal weights are also now renormalized (arguably the better behavior, but it's a scope expansion beyond the single-signal case the linked issue described).

- 🟡 `backend/src/hiresense/ingestion/domain/embedding_text/embed_profile_cached.py:31-34` (via `semantic_pre_ranker.py`) — the shared helper embeds `text = profile_text(skills, summary)` unstripped (only `text.strip()` is used for the emptiness check), whereas the pre-PR `SemanticPreRanker._get_profile_embedding` stripped the text before the actual `embed()` call: `text = f"...".strip()`. So a profile with an empty skills list now sends `"\n{summary}"` (leading newline) to the embedding port instead of the previously stripped `"{summary}"`, for the pre-ranker path specifically. This unifies onto `semantic_scoring_service`'s pre-existing (also-unstripped) behavior, which is consistent with the #161 goal, but it's an unacknowledged change to the literal embedded text for the pre-ranker — unlikely to shift rankings given typical embedding-model whitespace robustness, but worth a one-line mention in the PR description since it wasn't called out.

**Verified**
- Hand-computed `combine_fit_score` for skill-only, semantic-only, mixed present/absent, two-signal-default, and zero-weight cases against both the implementation and the new test assertions.
- Traced `IngestionOrchestrator.run()`'s single-flight guard/cooldown gate statement-by-statement for atomicity and correct release ordering (no `await` between check-and-claim; `_last_run_at` set before `finally` releases `_run_in_flight`).
- Diffed the pre-PR `semantic_pre_ranker._profile_key` (uncapped) against `semantic_scoring_service._profile_key` (capped) to confirm the claimed cache-key/embedding-text drift was real before this PR, and that the new unified `profile_key`/`profile_text` closes it.
- Confirmed the `job_text` extraction from `job_embedding_indexer.py` and `semantic_scoring_service.py` is byte-identical to both pre-PR versions (pure dedup, no behavior change).
- Grepped all callers of `combine_fit_score` (`ingestion/api/routes.py:300,441`, `semantic_pre_ranker.py:171`) — unaffected by the change under default weights.
- Grepped for lingering references to the removed private helpers repo-wide and found the untouched duplicate in `embedding_backfill_service.py` (see Findings).
- Checked `test_semantic_pre_ranker*.py` / `test_semantic_scoring_service.py` / `test_job_embedding_indexer.py` only exercise public methods, so the refactor doesn't require changes there.
- Confirmed `IngestionOrchestrator` is wired as a single bootstrap-time instance (`bootstrap/ingestion.py`), so the in-flight guard is scoped correctly (one instance per app process, not per-request).

Review by Claude Code

PR #197, opened 2026-07-20
fix: LLM reliability and silent-failure hardening

StevSant:
## LLM reliability and silent-failure hardening

A batch of four related reliability fixes: LLM calls can no longer hang forever, and three background/artifact paths no longer swallow failures into fake successes or invisible log lines.

### Per-issue summary

**#139 — Add a per-call timeout to LLM completions.** `LangChainLLMAdapter.generate()` (backing every `complete()`) now wraps the provider `ainvoke` in `asyncio.wait_for` with a configurable `llm_timeout` (default 60s), threaded from settings through `FeatureConfiguredLLMAdapter`. On expiry it raises the new `LLMTimeoutError`, mapped by a global exception handler to **504** so a stalled provider connection surfaces cleanly instead of tying up the async worker. Added `LLM_TIMEOUT` to `.env.example`.

**#142 — CV optimizer no longer fakes success.** `CVOptimizer.optimize()` used to wrap the whole flow in `except Exception` and, on any failure, return an `OptimizationResult` whose `optimized_tex == original_tex` — persisted as a 201, so the user believed their CV was tailored when it was byte-identical. It now raises `OptimizationError` on failure (mapped to **503**) and re-raises `LLMTimeoutError` untouched (504). A successful call that suggests no edits still returns identical text but with an empty change list, keeping "no changes" distinguishable from "failed".

**#147 — Interview prep no longer persists a placeholder as real prep.** `InterviewPrepService.prepare()` ended its LLM branch in a bare `except Exception` that returned a benign `InterviewPrep` (`negotiation_points=["Interview preparation is temporarily unavailable"]`) with no logging, which `ArtifactService` then persisted. It now logs with `logger.exception(...)`, re-raises `LLMTimeoutError` (504), and raises `InterviewPrepError` on other failures (mapped to **503**) so nothing bogus is persisted. The LLM-not-configured path (`llm is None`) still degrades gracefully and is unchanged.

**#163 — Observability on swallowed automation failures.** Added a `DomainMetrics.automation_failures_total` counter, incremented on the log-and-continue paths: scheduler job failure, inbox IMAP fetch, autohunt rerank, autohunt prune, and per-job autopilot drafting (labelled by `component`). Also added bounded retry with exponential backoff to the IMAP fetch for transient connection errors (`IMAP4.abort` / `OSError`), configurable via `IMAP_MAX_RETRIES` / `IMAP_RETRY_BASE_DELAY`.

Closes #139
Closes #142
Closes #147
Closes #163

[...]

PR #202, opened 2026-07-20
fix(outreach): SMTP/IMAP timeouts, TLS enforcement, CRLF header validation

### LINKED ISSUE #140: Reliability: Add timeouts to SMTP/IMAP connections (thread-pool exhaustion)
**Severity:** Medium
**Location:** `backend/src/hiresense/adapters/smtp_email_sender.py:39`; `inbox/infrastructure/imap_inbox_reader.py:47-49`
**Confidence:** High

## Description
`smtplib.SMTP(host, port)` and `imaplib.IMAP4_SSL/IMAP4(host, port)` are created with no `timeout`. Both run inside `asyncio.to_thread`, so a hung/black-holed mail server never times out the worker thread; repeated scheduled sends/scans can pin and exhaust the default thread pool, degrading the whole app. No `smtp_timeout`/`imap_timeout` exists in config.

## Proposed solution
Add configurable `smtp_timeout`/`imap_timeout` (default ~30s), pass to the constructors, document in `.env.example`.

## Acceptance criteria
- [ ] Both SMTP and IMAP connections use a configurable timeout
- [ ] `.env.example` updated

**Estimated effort:** S

### LINKED ISSUE #148: Security: Reject CRLF and validate to/subject in SmtpEmailSender (defense-in-depth)
**Severity:** Low
**Location:** `backend/src/hiresense/adapters/smtp_email_sender.py:34-44`
**Confidence:** Low

## Description
`mime["To"]`/`mime["Subject"]` are set from caller-controlled strings. The code uses the modern `email.message.EmailMessage` (policy=default), which refolds/validates headers and largely blocks CRLF header injection — so this is defense-in-depth, not an open hole. Still worth an explicit guard given the outreach relay feeds it untrusted values.

## Proposed solution
Reject values containing `\r`/`\n` before assignment (or validate via `EmailStr`), set recipients explicitly via `to_addrs`.

## Acceptance criteria
- [ ] Newline-bearing `to`/`subject` rejected
- [ ] Test asserts CRLF rejection

**Estimated effort:** S
**References:** CWE-93

### LINKED ISSUE #149: Security: Enforce TLS for SMTP/IMAP or refuse plaintext credential auth
**Severity:** Low
**Location:** `backend/src/hiresense/adapters/smtp_email_sender.py:39-43`; `inbox/infrastructure/imap_inbox_reader.py:46-53`; `config/groups/outreach.py:19-43`
**Confidence:** Medium

## Description
With `smtp_use_tls=false` the code calls `smtp.login()` over plaintext (credentials on the wire); with `imap_use_ssl=false` IMAP login is likewise plaintext. No explicit `ssl.SSLContext` is passed, so hardening relies on Python defaults, and no guardrail prevents a plaintext-auth misconfiguration.

## Proposed solution
Refuse to send login credentials over a non-TLS channel (or warn loudly at startup like the auth-secret path); pass an explicit verified `SSLContext`.

## Acceptance criteria
- [ ] No credential login over plaintext (or a loud startup warning)
- [ ] Explicit SSLContext used

**Estimated effort:** S

StevSant:
Hardens the SMTP/IMAP email surface in the outreach + inbox paths. Three related issues, one commit each.

### #140 — SMTP/IMAP connection timeouts (reliability, P2)
`smtplib.SMTP` and `imaplib.IMAP4/IMAP4_SSL` were constructed with no `timeout`. Both run inside `asyncio.to_thread`, so a hung/black-holed mail server could pin worker threads indefinitely and exhaust the default thread pool. Added configurable `smtp_timeout`/`imap_timeout` (default 30s), threaded through the `SmtpEmailSender` and `ImapInboxReader` constructors and their bootstrap wiring, documented `SMTP_TIMEOUT`/`IMAP_TIMEOUT` in `.env.example`.

### #149 — Enforce TLS / refuse plaintext credential auth (security, P3)
`smtp.login()` ran even with `SMTP_USE_TLS=false`, and IMAP logged in over a plaintext `IMAP4` socket with `IMAP_USE_SSL=false`, putting credentials on the wire. Neither passed an explicit `SSLContext` (TLS relied on library defaults, which do **not** verify certs). Now secure by default:
- SMTP refuses to `login()` over a non-TLS channel (raises `EmailUnavailableError` → 503); the best-effort IMAP scan logs a warning and no-ops.
- An explicit `ssl.create_default_context()` (cert + hostname verification) is passed to `STARTTLS` / `IMAP4_SSL`.
- New `SMTP_ALLOW_INSECURE` / `IMAP_ALLOW_INSECURE` flags (default `false`) are the single explicit escape hatch for a trusted local/dev server (e.g. mailhog): they permit plaintext auth and disable TLS verification.
- A server with no credentials (blank username) still sends over plaintext, so mailhog-style dev needs **no** flag.

### #148 — Reject CRLF in to/subject (security defense-in-depth, P3, CWE-93)
`EmailMessage` (policy=default) already refolds/validates headers, so this is defense-in-depth. Added an explicit guard rejecting `\r`/`\n` in `to`/`subject` (raises `ValueError`) before the message is built, and set the envelope recipient explicitly via `send_message(mime, to_addrs=[message.to])` instead of letting smtplib derive it from the untrusted `To` header.

## Closes
Closes #140
Closes #149
Closes #148

## Decisions taken
- **#149 secure-by-default with one opt-out flag per protocol.** Rather than separate flags for "allow plaintext auth" and "skip cert verification", each protocol gets a single `*_ALLOW_INSECURE` flag (default `false`) covering both, matching the issue's "opt-out defaulting to secure" and keeping config surface small. This flips prior behavior for TLS-enabled setups from *no cert verification* to *verified*; real mail providers (Gmail/Outlook/SES) are unaffected, self-signed/dev servers use the flag. `APP_MODE=local` is intentionally **not** auto-flipped: mailhog uses no auth, so the plaintext-auth guard never fires there.
- **#149 refuse-at-send, not startup warning.** The issue offered "refuse … or warn loudly at startup"; chose to refuse at send time (stronger, and the credentials aren't known to be insecure until a send is actually attempted).
- **#148 raises `ValueError` from the adapter (HTTP-agnostic).** The adapter stays free of HTTP concerns; a CRLF backstop maps to the route's generic `ValueError` arm. This path is a defense-in-depth backstop — the primary guard is the request schema (`to` is `EmailStr`, rejected at 422; `subject` is the realistic vector).

## Verification
- `uv run ruff check .` → all checks passed.
- `uv run ruff format --check` on all touched files → clean.
- `uv run python -m pytest` → full suite green except the documented environmental `test_autopilot_pipeline_settings_defaults` failure (local `.env` `AUTOPILOT_PIPELINE_ENABLED=true`; passes in CI). 12 new SMTP/IMAP tests cover timeouts, plaintext-auth refusal + insecure opt-in, verified-vs-unverified SSLContext, CRLF rejection, and explicit envelope recipient.

## Repo-trap checks
Config values added to `config/groups/outreach.py` + `.env.example` (no hardcoding); `SmtpEmailSender`/`ImapInboxReader` re-exports unchanged; domain layer untouched (guards live in the `adapters/` + `inbox/infrastructure/` layers, no framework imports leaked into `domain/`); one function/class per file preserved.

## Coordination
Expect **trivial** conflicts with sibling PR #197 (`fix/llm-reliability-batch`) on `config/groups/outreach.py` and `.env.example` — both branches add fields to `OutreachSettings`. #197 adds IMAP *retry* settings; this PR adds *timeout* + *TLS* settings. The additions are disjoint (different fields), so resolution is a straight both-keep. No overlap in the retry vs. connection-hardening logic.


StevSant:
## Self-review (critical)

Reviewed for correctness, security, and repo conventions. No blockers; a few things reviewers should weigh.

### Security / behavior changes worth a deliberate sign-off
- **Cert verification is now ON by default (behavior change).** `smtp_email_sender.py:73-78` and `imap_inbox_reader.py:_ssl_context` build `ssl.create_default_context()` (verify + hostname check) where the previous code passed no context (library default = *no* verification). Real providers (Gmail/Outlook/SES) are unaffected, but an existing deployment terminating TLS with a **self-signed** cert will start failing until it sets `SMTP_ALLOW_INSECURE`/`IMAP_ALLOW_INSECURE=true`. This is intended (see PR "Decisions taken") but is the one change an operator could feel on upgrade.
- **`ALLOW_INSECURE` disables verification (`CERT_NONE`).** `smtp_email_sender.py:75-77`. This is a deliberate, default-`false`, per-protocol escape hatch for trusted dev servers, matching the issue's "opt-out defaulting to secure." The linter/security hook flags `CERT_NONE` generically; it is gated behind explicit opt-in and documented in `.env.example`. Acceptable, but reviewers who want zero MITM surface even in dev could instead point `SSL_CERT_FILE` at a dev CA — noted, not changed.

### Correctness notes
- **`to_addrs=[message.to]` assumes a single recipient.** `smtp_email_sender.py:63`. Safe today: `EmailMessage.to` is a single `str` and the API schema types it `EmailStr` (`outreach/api/schemas.py`), so a comma-joined multi-recipient string can't reach here. If a future caller passes `"a@x.com, b@y.com"` in `to`, it would be treated as one envelope address. Left as-is to match the current contract; worth a comment if multi-recipient is ever added.
- **CRLF guard raises `ValueError` → HTTP 404, not 400.** `smtp_email_sender.py:70-71`. The adapter stays HTTP-agnostic, so a CRLF backstop maps to the route's generic `ValueError` arm (`outreach/api/routes.py:83`), which returns 404. This path is a defense-in-depth backstop only — the primary guard is `to: EmailStr` (422 at request parse) and `subject` refolding by `EmailMessage`. A dedicated error type mapped to 400 would be more correct but is out of scope for an S-sized defense-in-depth issue.
- **IMAP plaintext guard keys on a non-empty username.** `imap_inbox_reader.py:_fetch`. If `IMAP_USE_SSL=false` **and** username is blank, the guard doesn't fire and `client.login("", "")` still runs over plaintext — but an unauthenticated IMAP scan is nonsensical and the outer `except` in `fetch_unseen` degrades it to `[]`. Acceptable.

### Minor
- `_ssl_context()` builds a fresh context per send/fetch (`smtp_email_sender.py:73`). Negligible for low-volume outreach; could be memoized if volume grows.
- `EmailUnavailableError` for the plaintext-auth refusal surfaces as the generic 503 "not configured" message (`outreach/api/routes.py:85-88`). Semantically it's a *misconfiguration* rather than "unset," but 503 is the right class and reusing the type keeps wiring minimal.

### Conventions
Config in `config/groups/outreach.py` + `.env.example` (no hardcoded values); domain layer untouched (guards live in `adapters/` and `inbox/infrastructure/`); one class/function per file; re-exports unchanged. `ruff check`, `ruff format --check`, and the full pytest suite are green (only the pre-existing environmental autopilot-default failure remains).

StevSant (commented):
## Code review — #202

**Done well**
- `backend/src/hiresense/adapters/smtp_email_sender.py:29-33` and `backend/src/hiresense/inbox/infrastructure/imap_inbox_reader.py:44-49` — the plaintext-login refusal check runs *before* any socket is opened (verified by `assert created == []` in `test_send_refuses_login_over_plaintext` / `test_fetch_refuses_login_over_plaintext`), so no partial connection or credential exposure on the refused path.
- `_ssl_context()` in both files sets `context.check_hostname = False` before `context.verify_mode = ssl.CERT_NONE` — the correct order (Python's `ssl` module raises `ValueError` if you flip it), and defaults to a verified `ssl.create_default_context()` otherwise.
- STARTTLS failure is fail-closed by construction: `smtp.starttls(context=...)` is called unconditionally when `use_tls=True`, so a stripped/absent STARTTLS advertisement raises `SMTPNotSupportedError` and aborts the send rather than silently downgrading.
- CRLF guard (`_reject_header_injection`) checks `\r` and `\n` independently (not just the `\r\n` pair) and covers both header-bound fields on `EmailMessage` (`to`, `subject`) — confirmed `kernel/email_message.py` has no other header-bound fields (no reply-to/custom-headers path exists), so nothing is missed.
- Explicit envelope recipient (`to_addrs=[message.to]`) closes the gap where smtplib would otherwise derive RCPT TO from the attacker-influenced `To` header.
- `timeout` is a required (no-default) constructor kwarg on both `SmtpEmailSender` and `ImapInboxReader`, so every call site was forced to be updated — grepped all constructions (`bootstrap/outreach.py`, `bootstrap/notifications.py`, `bootstrap/inbox.py`, plus tests) and all pass it explicitly; no caller left with an unbounded socket.
- Timeout failures don't produce a false "sent" state: in `OutreachService.send()` the `OutreachEvent` is only persisted *after* the awaited `sender.send()` call returns, so a timeout/`SMTPException` propagates before any record is written — no silent "half-sent" bookkeeping. `NotificationService._safe_send` independently swallows send errors (logged, returns `False`) for the best-effort digest/failure paths, while `send_test()` still lets errors surface as 503 — both pre-existing patterns, correctly preserved.
- Config additions (`smtp_timeout`, `smtp_allow_insecure`, `imap_timeout`, `imap_allow_insecure`) go through `config/groups/outreach.py` + `.env.example` with explanatory comments, per repo convention — no hardcoded values.
- Good test coverage: 12 new tests hit both the secure-default and insecure-opt-in paths for TLS/SSL context, timeout plumbing, plaintext-login refusal, and CRLF rejection on `to`/`subject`.

- 🟡 `backend/src/hiresense/outreach/api/routes.py:79-80` — the CRLF-guard's `ValueError` (raised from `_reject_header_injection` deep in the sender) is caught by the route's generic `except ValueError as exc: raise HTTPException(status_code=404, ...)` arm, so a malformed `subject` returns `404 Not Found` instead of a `400 Bad Request`. Functionally safe (the send is still refused, nothing leaks), but the status code is misleading for a validation failure — a caller could read 404 as "application not found" rather than "your subject is invalid." This is a documented, deliberate reuse of the existing generic-`ValueError`-as-404 arm per the PR description, so not blocking — just worth a follow-up to give `_reject_header_injection` its own exception type mapped to 400 if this route sees real traffic with untrusted subjects.
- 🟡 `backend/src/hiresense/config/groups/outreach.py:20-23,50-53` — `smtp_timeout`/`imap_timeout` are plain `float` fields with no lower-bound validation. A misconfigured `SMTP_TIMEOUT=0` (or negative) would pass a `0`/negative timeout straight into `socket`-level operations, which for `timeout=0` puts the socket in non-blocking mode and will likely break sends/reads outright rather than degrading gracefully. Low real-world likelihood (operator-set config, not attacker-controlled), but a `Field(gt=0)` constraint would make the failure mode a clear startup validation error instead of a confusing runtime one.

**Verified**
- Full diff of all 11 changed files (`git diff origin/main...origin/fix/outreach-email-batch`), including `.env.example`, both bootstrap wiring changes, and the config group.
- Traced `EmailMessage` (kernel model: `to`/`subject`/`body` only, no custom headers) to its two production construction sites (`outreach_service.py:127`, `notification_service.py:74`) and confirmed `subject`/`message` in `SendRequest` are plain `str` (the realistic CRLF vector — `to` is `EmailStr`, already 422-rejected upstream for CRLF).
- Grepped every `SmtpEmailSender(`/`ImapInboxReader(` call site across `backend/src` and `backend/tests` to confirm none was missed when adding the required `timeout` kwarg.
- Confirmed `ssl.create_default_context()` verified-vs-unverified branching order is correct in both adapters, and that `starttls()`/`IMAP4_SSL(...)` fail closed rather than silently downgrading.
- Did not execute the test suite myself (read-only review environment, no branch checkout available); relied on code inspection plus the PR's stated `ruff`/`pytest` verification output.

**Verdict:** LGTM ✅

Review by Claude Code

PR #208, opened 2026-07-20
fix(autohunt): autopilot pipeline idempotency under concurrent runs

### LINKED ISSUE #136: Architecture: Make the autopilot pipeline idempotent (TOCTOU duplicates)
**Severity:** Medium
**Location:** `backend/src/hiresense/autopilot/domain/autopilot_pipeline_service.py:40-54`; `autopilot/infrastructure/draft_repository.py:39-42`; `autopilot/infrastructure/autopilot_draft_orm.py`
**Confidence:** Medium

## Description
Dedup relies on a read-then-write `exists_for_job(job_id)` with no DB uniqueness on `job_id`. APScheduler `max_instances=1` only prevents the scheduled job overlapping itself; a manual `POST /autopilot/run` firing concurrently with the scheduled run is a TOCTOU window where both pass the existence check and each calls `create_from_ingested`, producing duplicate applications plus duplicate LLM-generated artifacts (wasted spend, dirty tracking data).

## Proposed solution
Add a unique constraint on `autopilot_drafts.job_id` (+ migration) and let the insert be the idempotency guard (catch `IntegrityError` → skip), or serialize runs with a DB advisory lock. Add a concurrent-run regression test.

## Acceptance criteria
- [ ] DB uniqueness (or advisory lock) prevents duplicate drafts for the same job
- [ ] Migration added and applied
- [ ] Test simulates concurrent scheduled+manual run and asserts one draft

**Estimated effort:** M

StevSant:
# What

Makes the autopilot pipeline idempotent under concurrent runs. Two runs racing on the same job (scheduled run vs manual run-now, or two processes) could both pass the `exists_for_job` pre-check (TOCTOU) and draft the same job twice — a duplicate application plus duplicate LLM-generated artifacts.

Closes #136

# How

The draft row itself is now the idempotency guard:

- **Migration 038**: collapses any pre-existing duplicate `autopilot_drafts` rows (keeping the earliest per `job_id`), then rebuilds the `job_id` index as **UNIQUE**.
- **Claim-first flow**: the pipeline inserts a `PENDING` reservation row *before* any expensive drafting. The loser of a concurrent race gets an `IntegrityError` from the unique constraint, `claim()` returns `None`, and the run skips the job — no duplicate application, no wasted LLM spend.
- **Port surface**: `claim`/`finalize` added to the `DraftRepository` protocol; `finalize` updates the reserved row with the drafting outcome. `DraftStatus.PENDING` marks an in-flight reservation.
- The old `exists_for_job` check remains as a cheap pre-filter for jobs drafted in *prior* runs; the DB constraint is the real guard for *concurrent* ones.

# Test plan

- New `tests/integration/test_autopilot_concurrency.py`: two independent service instances (defeating the in-process guard) share one file-backed SQLite DB and race on the same job — asserts exactly **one** draft row and **one** drafter invocation.
- New repository tests for `claim` (reserve, then reject duplicate) and `finalize` (in-place update).
- Endpoint/unit fakes updated to the new port surface; endpoint test now waits for finalization (not merely row presence) before asserting.
- Full suite: 1371 passed locally (single failure is the known-environmental `test_autopilot_pipeline_settings_defaults`, green in CI). `ruff format`/`ruff check` clean. `alembic heads` = single head `038`.

# Notes

- **After merge, upgrade the dev DB**: `uv run python -m alembic upgrade head` (migration 038), or the app will 500 on autopilot endpoints once the ORM expects the unique index.
- Residual edge (documented, accepted): a hard process crash *mid-draft* leaves a `PENDING` reservation that blocks re-drafting that job (drafter exceptions are already handled and finalize as `FAILED`; only a kill -9-class crash leaves PENDING). A follow-up could add a staleness sweep if this shows up in practice.


StevSant:
**Self-review notes** (controller-verified after the implementing agent stalled):

- Migration 038 reviewed: dedupes pre-existing duplicates (earliest per `job_id` kept, `ROW_NUMBER` over `created_at, id`) *before* recreating the index as unique — safe on a dirty dev DB. `down_revision = "037"` matches the current single head; no other open PR adds a migration.
- `claim()` maps only `IntegrityError` to `None`; every `SqlRepository` helper opens its own session, so the failed insert cannot poison a shared session.
- Caught and fixed in this branch: the endpoint-test fake `_Repo` had not been updated to the new `claim`/`finalize` port surface. The background run crashed before reaching the gated drafter, and `await drafter.entered.wait()` hung the suite indefinitely — this was why local full-suite runs appeared to hang. Fake updated; both endpoint tests pass in 0.4s.
- Residual edge (accepted, in PR notes): a hard process crash mid-draft leaves a `PENDING` reservation blocking that job; drafter exceptions already finalize as `FAILED`. A staleness sweep is a possible follow-up.

StevSant (commented):
## Code review — #208

**Done well**
- The claim-first design genuinely closes the TOCTOU: the reservation insert happens *before* any LLM/drafting work (`autopilot_pipeline_service.py::_draft_one`), so the loser of a race never spends LLM tokens or creates a duplicate application — it just gets `None` back from `claim()` and is folded into `skipped`, not `created`.
- `tests/integration/test_autopilot_concurrency.py` is a real proof, not a tautology: it uses two independent `AutopilotPipelineService` instances (defeating the in-process `_running` flag) sharing one file-backed SQLite DB with a busy timeout, widens the window with `asyncio.sleep(0.05)` inside the fake drafter, and asserts exactly one row + one drafter invocation. This would fail without the unique constraint.
- Migration 038's dedupe-then-constrain approach (delete extras via `ROW_NUMBER() OVER (PARTITION BY job_id ORDER BY created_at, id)`, then rebuild the existing `ix_autopilot_drafts_job_id` index as unique) is correct and safe: `created_at` is `nullable=False` on `autopilot_drafts` (backend/src/hiresense/autopilot/infrastructure/autopilot_draft_orm.py), so there's no NULL-ordering edge case, and the index name matches the one originally created in migration 035, so `alembic check` (run in CI) will not flag drift.
- `finalize()`'s `IntegrityError`/vanished-row `RuntimeError` path is correctly marked `pragma: no cover` — there is no delete path anywhere on `autopilot_drafts`, so a claimed row can't disappear before finalize.
- Test updates are proportionate: `test_pipeline_service.py`'s `_Repo` fake now models `claim`/`finalize` faithfully (rejecting a second claim for the same `job_id`), and the endpoint test (`test_autopilot_endpoints.py`) correctly waits for the row to leave `PENDING` before asserting completion, instead of just checking row presence — a real fix given the new intermediate state.

- 🟡 `frontend/src/app/core/models/autopilot.model.ts:7` — `status: 'drafted' | 'partial' | 'failed'` was not updated for the new `pending` value. `GET /autopilot/drafts` (`backend/src/hiresense/autopilot/api/routes.py:36-41`) returns `provider.get_repo().list(limit)` with no status filter, so a client polling the list while a run is mid-flight can now legitimately receive a draft with `status: "pending"`. The Angular template renders it as `class="status status--pending"` (`frontend/src/app/pages/autopilot/drafts/drafts.component.html:12`), but `drafts.component.scss` only defines `&--drafted`, `&--partial`, `&--failed` (lines 16-24), so a pending row would show unstyled (default text color) with a TS type that doesn't account for the value it can actually receive. Cosmetic only — the window is short and nothing crashes — but the frontend contract is out of sync with the backend's new state.
- 🟡 `backend/src/hiresense/autopilot/infrastructure/draft_repository.py:41-46` — `claim()` catches bare `IntegrityError`, not specifically the unique-constraint violation, and returns `None` (interpreted as "already claimed"). In practice this is safe today: `job_id` and `status` are non-nullable but always supplied by the domain model/pipeline, so the unique index is the only constraint that can fail on this insert. Still, if a future column addition introduces another NOT NULL/FK constraint on `autopilot_drafts`, a distinct integrity failure would be silently swallowed as a normal "lost the race" case instead of surfacing as an error. Consider narrowing to the specific constraint (e.g. checking `orig` args) or adding a comment noting the assumption so it's revisited if the schema grows.
- 🟡 `backend/src/hiresense/autopilot/infrastructure/draft_repository.py:24-25` / `domain/ports/draft_repository.py:10` — `add()` remains on both the port and the impl, but after this change no production code path calls it (only `tests/integration/test_autopilot_repository.py::test_add_list_exists` exercises it directly). Since `add()` bypasses the unique-reservation guard entirely, keeping it in the public port invites a future caller to reintroduce the exact TOCTOU this PR fixes. Consider removing it from the port (or renaming/privatizing) now that `claim`/`finalize` are the sole write path, or at minimum note in its docstring that it must not be used by the pipeline.

**Verified**
- Migration chain: 035 → 036 → 037 → 038 is linear with a single head; `down_revision = "037"` matches 037's `revision = "037"`.
- No other write path to `autopilot_drafts` bypasses `claim()`/`finalize()` — grepped the full PR-branch tree for `DraftRepositoryImpl`/`.claim(`/`.finalize(`/`repo.add(`; the only `.add(` hits belong to unrelated repos (`autohunt_service.py`, `outreach_service.py`) and test files for those domains.
- `SqlRepository._insert`/`_update_by_pk` each open a fresh session via `self._session_factory()` per call (standard sessionmaker pattern), so an `IntegrityError` raised inside `claim()`'s `_insert` doesn't poison a shared/reused session for subsequent repository calls.
- CI (`​.github/workflows/ci.yml:66-70`) applies this migration against a real Postgres service container and runs `alembic check` afterward, so the raw SQL and the index-name match against the ORM's `unique=True` will be validated on every PR, not just locally.
- The documented residual risk (a hard crash mid-draft leaves a blocking `PENDING` row, since `exists_for_job` treats any row — regardless of status — as "already drafted") is real and accurately described in the PR body; there's no admin/ops tooling to clear a stuck `PENDING` row today, but this is an explicitly accepted, pre-existing-shaped limitation rather than a regression, so not counted as a new finding.
- `PipelineResult` accounting (`created`/`skipped`) correctly treats a lost-race (`None`) result as skipped, not created, and this is asserted by the concurrency test.

**Verdict:** LGTM ✅

Review by Claude Code

PR #224, opened 2026-08-09
fix(profile): repair apply profile save, drop the redundant Account tab

StevSant:
Fixes two user-reported problems on the Profile page.

## 1. Apply profile save hung and silently discarded answers

Filling in the Apply Assist form left the button stuck on "Saving…" and nothing persisted. Two independent defects:

- **The input `effect()` clobbered in-progress edits.** It re-synced `form` *and* `baseline` on every profile identity change, with no dirty check. Any background refetch mid-edit wiped what the user had typed and reset `isDirty()` to false — after which `save()` early-returns and **no request is ever fired**. This is the "didn't save" half.
- **`saving` could never reset.** It was only cleared inside `next`/`error`, but `takeUntilDestroyed` tears the stream down via *unsubscribe*, which calls neither. Moved to `finalize()`, which covers all three paths. This is the stuck-button half.

Also fixed an adjacent cache bug: the apply profile is one-per-person and the backend writes it to every language row, but `ProfileService` keyed the response by its own language, leaving other languages stale.

## 2. The Account tab was redundant

It contained only a read-only username/role pair, a note saying both are env-managed, and a Sign Out button the sidebar already has. It also nested its own `<h1>Account</h1>` inside the page's `<h1>My Profile</h1>`, breaking heading order.

Removed it, and moved the one piece of information with no other home — who you're signed in as — into the sidebar footer above Sign Out. `/dashboard/account` still redirects to the profile page, so existing links resolve.

## Test plan

- [x] 4 new regression tests on `apply-profile-card`; **2 verified failing against the old code** before the fix (form clobber, stuck `saving`)
- [x] 3 new tests on the sidebar identity block (username + role, admin badge, absent when no session)
- [x] Full frontend suite: **506 passed / 87 files**
- [x] `npx ng lint` clean
- [x] `prettier --check` clean on touched files
- [x] `npm run build` succeeds


StevSant (commented):
## Code review — #224

**Correct / well handled**
- `apply-profile-card.component.ts` — the `effect()` now gates the input re-sync on `untracked(() => this.isDirty())`, correctly restoring the "hold local edits until save" behavior without turning `form`/`baseline` reads into reactive dependencies (would have caused the effect to re-fire on every keystroke). This fixes the real bug: a background refetch minting a new `profile` identity used to always clobber `form`/`baseline`, silently zeroing `isDirty()` so `save()`'s early-return meant the request never fired.
- Moving `saving.set(false)` into `finalize()` correctly covers the `takeUntilDestroyed` teardown-via-unsubscribe path, which calls neither `next` nor `error` — the previous code really could strand the button on "Saving…" forever. New test `clears the saving flag when the request is torn down without emitting` (using `NEVER` + `fixture.destroy()`) directly exercises this path.
- `profile.service.ts` `setApplyProfile` — fanning `apply_profile` out across every cached language row (instead of only the response's own language) matches the stated one-per-person backend model; other cached languages keep their own fields via the spread and only get `apply_profile` overwritten.
- Account tab removal is clean: `/dashboard/account` redirect in `app.routes.ts` still exists (`redirectTo: 'profile'`), no leftover imports/selectors/route references to `AccountComponent` or `app-account` anywhere on the branch, and the sidebar footer gains the one piece of info (`username`/`role`) that had no other home, sourced from new `AuthService.username`/`user` readonly signals.
- Good regression coverage: 4 new tests on `apply-profile-card` (clobber-on-refetch, re-sync-when-clean, stuck-saving-on-teardown, error-path-clears-saving) and 3 on the dashboard sidebar identity block (username+role, admin badge styling, absent-when-no-session).

- 🟡 `frontend/src/app/pages/profile/profile.component.html:4` — the page subtitle still reads "Your CV, personal details, cover letters, **and account** in one place." after the Account tab was removed; leftover copy referencing a section that no longer exists.

**Verified**
- Full diff (`git diff origin/main...origin/feat/profile-analytics-restructure`) — 15 files, no backend changes, so no ruff/Alembic/registry concerns apply.
- Confirmed no dangling references to the deleted `AccountComponent`/`app-account`/`'account'` pageTab anywhere in `frontend/src` on the PR branch (`git grep` on the remote branch tree), and the `/dashboard/account` → `/dashboard/profile` redirect route is intact.
- Traced the `ApplyProfileCardComponent` effect/save interaction end-to-end (constructor effect, `save()`, `toPayload`, `ProfileService.setApplyProfile`) — the fix genuinely closes both defects described in the PR body; no other silent-swallow paths introduced.
- No i18n framework is used in this frontend (plain English strings throughout), so no i18n-convention gap here.
- Heading order: only one `<h1>` (`My Profile`) remains in the profile page template; the removed component's nested `<h1>Account</h1>` is gone.

**Verdict:** LGTM ✅

Review by Claude Code

PR #225, opened 2026-08-09
fix(applications): stop truncating the application list; extract a shared paginator

StevSant:
Item 5 of the UI restructure. Independent of #224 — branched off `main`.

## The bug

`ApplicationsService.list()` took no arguments, so it always fetched the backend's default page (100 rows) and threw away the `X-Total-Count` header the endpoint already sets (`applications/api/routes.py:112`). Nothing in the frontend read that header anywhere.

Beyond dropping rows, `applications.component.ts` computes the **status-tab badge counts** from that truncated set — so past 100 applications the numbers on screen were simply wrong. The same `list()` also backs the outreach application picker and the interview prep list.

## Why load-all rather than server-side paging

The applications list, the outreach picker and the cover letter library all search, sort and count **client-side**. Paginating server-side would have made those operations cover only the current page — a worse regression than the one being fixed — and moving them server-side is the backend `Page[T]` work that's deliberately deferred to separate issues.

So `listAll()` walks to the last page, and the paginator runs over the client-side filtered rows to keep the DOM bounded. The walk stops at `environment.listMaxItems` and the page says "showing the first N of M" rather than implying completeness. Termination is guarded three ways (covered offset, empty page, ceiling) so a wrong server total can't spin an infinite request loop.

## Changes

- **Shared paginator** — moved `pages/ingestion/components/pagination/` → `core/components/paginator/` (git history preserved); it was already generic, which is why `opportunities` reached across a feature boundary to import it. Added a `pageSizeOptions` input since the hardcoded 20/50/100 doesn't match every endpoint's cap.
- **`/dashboard/applications`** — full load + pager + correct badge counts.
- **Cover letter library** — same treatment.
- **`/dashboard/company/:name`** — was merging two fixed 100-row pages under a comment claiming "all jobs load in one page"; now walks both tabs. The comment is corrected.
- **`/dashboard/admin/usage`** — backend gained `count_recent` (sharing WHERE clauses with `list_recent` so listing and count can't drift) surfaced as `RecentCallsResponse.total`; frontend now uses its previously-unused `recentLimit` and a real offset. Filter/sort changes reset to page 1.
- **Portfolio card** — migrated onto the shared paginator, deleting its duplicated offset math.

## Verification

- Frontend: **520/520** tests pass, `ng lint` clean, production build succeeds, prettier clean on touched files.
- Backend: **1659 passed, 6 skipped**, `ruff format --check` and `ruff check` clean.
- The regression tests in `applications.service.spec.ts` were confirmed to **fail against the old single-page implementation** before the fix was restored.

One deliberate behavior change: the portfolio card's range indicator now reads `1–12 of 30` (page × pageSize) instead of `1–1 of 30` (rows actually returned). The shared formula is correct for real full-page responses and keeps the range stable while a page loads; its spec was updated with a note.


StevSant (commented):
## Code review — #225

**Correct / well handled**
- The actual bug is fixed at every layer: `applications.service.ts` drops the old argument-less `list()` in favor of `listPage`/`listAll`, `toPagedResult` reads `X-Total-Count` with a safe fallback, and `fetchAllPages` walks pages with three independent termination guards (covered offset, empty page, `maxItems` ceiling) — verified against `fetch-all-pages.spec.ts`, which explicitly tests a server total that never matches reality (`fetcherOver([1,2,3], 999)`) to confirm the empty-page guard prevents an infinite loop.
- Every previous consumer of the truncated `list()` was migrated: `applications.component.ts`, `cover-letter-library.component.ts`, `applications-prep-list.component.ts`, and `outreach.component.ts` all now call `listAll()` and unwrap `{ items, total }`. Grepped the whole branch tree for leftover `.list(` call sites — none found.
- Paginator boundary math is correct and tested: `showingFrom`/`showingTo` clamp correctly on page 1, last page (not evenly divisible), and empty (`total === 0`) in `paginator.component.spec.ts`; Prev/Next correctly disable at the edges and don't emit when disabled; page-size change resets to page 1 at every call site (`onPageSizeChange` → `page.set(1)`).
- `applications.component.ts`/`cover-letter-library.component.ts` client-side paging: `totalPages`/`currentPage` clamp so narrowing a filter can't strand the user past the last page — no off-by-one in the slice math (`(currentPage - 1) * pageSize`).
- Backend `count_recent` shares its WHERE-clause builder (`_calls_filters`) with `list_recent`, exactly as the PR claims — listing and count can't drift apart. New `test_usage_aggregator_count.py` covers the pass-through and the `days → since` translation.
- `admin-usage.component.ts`: the fix for filter/sort losing pagination context is real — the HTML diff shows `(click)="loadRecent()"` → `(click)="reloadRecent()"` and every `(sorted)="loadRecent()"` → `(sorted)="reloadRecent()"`, and `reloadRecent()` resets `callsPage` to 1 before reloading, matching the "stale limit inputs" gotcha this exact pattern caused before.
- Old `pages/ingestion/components/pagination/` → `core/components/paginator/` move is a clean git rename (confirmed via `--name-status`, R062–R100), and both remaining consumers (`ingestion.component`, `opportunities.component`) were updated to the new import/selector with no leftover references.
- Environment-driven config (`listPageSize`, `listMaxItems`) is properly added to all three environment files, not hardcoded in application code — consistent with the project's no-hardcoded-values convention.
- The portfolio card's `1–12 of 30` vs `1–1 of 30` behavior change is deliberate, documented in the PR body, and covered by an updated spec with an explanatory comment — not a regression.

- 🟠 `frontend/src/app/pages/company/company.component.ts:150-166` — `allJobsFor()` walks `queryJobs` with the default `rescore = true` on every page. Per the backend's own comment in `ingestion/api/routes.py:576-583`, `rescore=false` is meant for exactly this case ("pure reorder/pagination" that doesn't change the filter/tab), and `rescore=true` triggers a blocking LLM call (`llm_on_miss=true`) for every uncached job on every page. Before this PR the walk was capped at one fixed 100-row page per tab; now it walks every page for both `boards` and `portals`, so a company with, say, 500 jobs across sources can turn a "view this company" click into hundreds of blocking LLM scoring calls instead of the two capped ones today. Since this component already prefers the persisted score (`job.llm_score ?? job.match_score`) and only needs a stable sort, passing `rescore: false` (there is already a `rescore` param on `queryJobs`) would avoid the multiplication without changing correctness.
- 🟡 `frontend/src/app/pages/company/company.component.ts:159` — `Math.floor(offset / limit) + 1` assumes every intermediate page returned exactly `limit` rows. That holds under normal operation (the walk only continues while `loaded < total`, and pages before the last one are expected to be full), but if the server ever returns a short page before the total is reached, the page-number conversion silently mis-maps to the wrong `queryJobs` page. Low risk given current server behavior, but worth a comment or an assertion if it's meant to be a hard invariant.
- 🟡 `frontend/src/app/core/components/paginator/paginator.component.html` — the pagination controls aren't wrapped in a `<nav aria-label="Pagination">` (or similar) landmark, and there's no `aria-live` region around `.pagination-info`, so a screen-reader user isn't told the range/page changed after clicking Prev/Next. Not a regression (the old `pagination.component.html` had the same shape), but since this is now the shared, reused primitive across 6 call sites it's a good point to add it once.

**Verified**
- Read every changed file in the diff (43 files) via `git show origin/feat/shared-paginator:<path>` — none read from the working tree.
- Confirmed no orphaned old `pagination` directory (rename detected, not copy+delete).
- Confirmed `X-Total-Count` is already set by the backend on both `/applications` and `/applications/cover-letters` (untouched by this PR, as claimed) — `toPagedResult`'s fallback path is exercised for endpoints that haven't adopted the header.
- Confirmed hexagonal layering for the new `count_recent`: routes → `UsageAggregator` (domain) → `LLMUsageLogRepositoryPort` (port) → `LLMUsageLogRepository` (infra); no framework leakage into domain.
- Spot-checked backend admin schemas/routes/domain/port/infra for the new `total` field — all four layers consistent, no drift.

Review by Claude Code

PR #226, opened 2026-08-09
refactor(profile): replace the hand-rolled tab bar with routed hub tabs

StevSant:
Item 3 of the UI restructure. **Stacked on #224** — review that one first; the base will retarget to `main` automatically once #224 merges.

## Why

Profile was the last page running its own signal-driven tab bar. Its sections weren't deep-linkable, ignored back/forward, and loaded eagerly. It also forced a special case elsewhere:

- `dashboard.component.ts` suppressed the shared hub tab bar specifically for the profile hub
- `hubs.const.ts` carried a sentinel tab that existed only so `hubForUrl` could resolve the URL

Both are now gone — profile uses the same routed hub tabs as every other hub.

## Structure

Three lazily-loaded child routes: `/dashboard/profile/{cv,personal,cover-letters}` (the bare parent redirects to `cv`).

Nearly all of the old 257-line component was CV-specific (upload, LaTeX paste, per-language variants, translation, PDF download) and moves wholesale into the CV tab. The shell keeps only the header and the one-time bootstrap.

**One design consequence worth flagging:** routed children can't read a loading flag off a parent component, so `ProfileService` now owns a `loaded` signal. The shell bootstraps once and flips it; each tab derives its spinner from it. No tab refetches on navigation.

The stylesheet is split the same way — genuinely shared rules (loading state, buttons, alerts, `details-card`, empty state) live in a `_profile-tab-shared.scss` partial each tab `@use`s, since Angular scopes component styles. The `.page-tabs` / `.page-tab` / `.tab-panel` rules are deleted along with the markup they styled.

The "Upload a CV in the CV tab" hint in Personal details became a real `routerLink` instead of a `pageTab.set('cv')` button.

## Tests

The old `profile.component.spec.ts` tested CV logic through the page component; it's split to follow the code — shell bootstrap, CV tab behavior, Personal tab behavior — with a shared `makeProfile` fixture. Two specs asserted the removed special-casing and were updated to assert the new behavior:

- `dashboard.component.spec.ts` — profile now returns a hub instead of `null`
- `hub-for-url.spec.ts` — resolves `/dashboard/profile/cv` (the bare parent redirects, and the dashboard reads the post-redirect URL)

## Verification

**510/510** tests pass, `ng lint` clean, production build succeeds, prettier clean on touched files.


StevSant (commented):
## Code review — #226

**Correct / well handled**
- `app.routes.ts` — the three child routes are wired correctly: bare parent `{ path: '', redirectTo: 'cv', pathMatch: 'full' }` plus lazy `loadComponent` for `cv` / `personal` / `cover-letters`. `canActivate: [authGuard]` lives on the parent `dashboard` route and is inherited by all children, so no guard was dropped.
- `ProfileService.loaded` is a clean solution to the "routed children can't read a loading flag off the parent" problem; `profile-cv-tab.component.ts` and `profile-personal-tab.component.ts` both derive `initialLoading = computed(() => !this.profileService.loaded())` from it, and the shell's `ngOnInit` sets it exactly once regardless of success/failure path — no tab re-fetches on navigation.
- `hubForUrl`/`hubs.const.ts`/`dashboard.component.ts` — the sentinel-tab special case and the `id === 'profile'` suppression in `hubTabs` are both removed cleanly; `dashboard.component.ts:29` reads `router.url` (post-redirect at construction) / `urlAfterRedirects` (post-redirect on navigation), so `hubForUrl` correctly resolves `profile` for `/dashboard/profile/cv` and the shared `HubTabsComponent` (`role="tablist"`/`role="tab"`/`aria-selected`) now renders for Profile like every other hub — no a11y regression introduced by this PR (the tablist/tabpanel wiring gap in `hub-tabs.component.html` is pre-existing and shared by all hubs, not new here).
- Old 257-line `profile.component.ts`/`.html`/`.scss` fully gutted to a thin shell (header + one-time bootstrap + `<router-outlet>`); `.page-tabs`/`.page-tab`/`.tab-panel` CSS and the `pageTab` signal are gone with no orphaned references left behind.
- Test split mirrors the code split, with a shared `makeProfile` fixture; `dashboard.component.spec.ts` and `hub-for-url.spec.ts` were correctly updated to assert the new (non-special-cased) behavior instead of just deleting the old assertions.

- 🟠 `frontend/src/app/pages/profile/tabs/cover-letters-tab/profile-cover-letters-tab.component.ts` (imports: `[CoverLetterLibraryComponent, CoverLetterTemplatesComponent]`) + `profile-cover-letters-tab.component.html:24` — `<a routerLink="/dashboard/applications" class="btn-primary per-job-cta"> Go to Applications </a>` uses the `routerLink` directive, but `RouterLink` is not imported on this standalone component. In the old monolithic `profile.component.ts`, `RouterLink` was imported once and covered this exact anchor (see `origin/feat/profile-analytics-restructure:frontend/src/app/pages/profile/profile.component.html:612`). Since the attribute is used unbracketed (`routerLink="..."` not `[routerLink]="..."`), Angular's template compiler treats it as an inert static HTML attribute on a native `<a>` rather than flagging an unknown binding — so this builds and lints clean, but at runtime the directive never activates: the anchor gets no `href` and no click interception, so the "Go to Applications" CTA is a dead link (no navigation happens on click). This is exactly the kind of regression the PR's own test suite wouldn't catch, since `ProfileCoverLettersTabComponent` has no spec file. Fix: add `RouterLink` to both the `import` statement and the component's `imports` array.

**Verified**
- Diff scope: `git diff origin/feat/profile-analytics-restructure...origin/feat/profile-routed-tabs` (23 files, +1995/-1928).
- Read full new-branch content via `git show origin/feat/profile-routed-tabs:<path>` for `app.routes.ts`, `hubs.const.ts`, `hub-for-url.spec.ts`, `profile.service.ts`, `dashboard.component.ts`/`.spec.ts`, `profile.component.ts`/`.html`/`.spec.ts`, all three new tab components (`.ts`/`.html`) and their specs, `make-profile.ts`, and the shared `hub-tabs.component.html` (pre-existing, unmodified by this PR).
- Grepped every new/changed template for `routerLink` usage and cross-checked each against its component's `imports` array — only the cover-letters-tab anchor is missing the import.
- Confirmed `.page-tabs`/`.page-tab`/`.tab-panel` classes and the CV-tab-only `language-tabs` block (pre-existing, unrelated `role="tab"` usage) are correctly scoped — no leftover dead CSS.
- Confirmed the auth guard is inherited from the parent `dashboard` route (children add no `canActivate`, matching pre-PR behavior — no guard was lost).

Review by Claude Code

PR #227, opened 2026-08-09
refactor(analytics): split the page into routed insight tabs

StevSant:
Item 4 of the UI restructure. **Stacked on #226** (which is stacked on #224) — both touch `hubs.const.ts`, so they're sequenced rather than parallel.

## Why

The insights hub already rendered a one-item tab bar above this page, so the page did **not** need its own in-page bar — it needed the hub filled in.

More importantly, Market and Skill gap were buried inside a collapsed `<details class="analytics-context">`. The two sections most likely to answer "what should I learn next?" were the hardest to reach. They're now first-class tabs.

## Grouping

| Tab | Contents |
|---|---|
| **Pay** | comp benchmark + the market salary band (moved out of the buried Market block, next to the benchmark it belongs with) |
| **Fit** | search focus + skill gap |
| **Pipeline** | funnel + outcomes by source |
| **Market** | top skills, remote mix, postings per week |
| **Portfolio** | engagement rows |

## What tabs buy here (and what they don't)

The KPI strip stays **above** the tabs so the tiles summarise every section whichever tab is open. That constrains what can be deferred: `comp`, `funnel` and `focus` feed the tiles and must stay eager. Only `market`, `skillGap` and `engagement` actually load with their tab — so this is mostly a UX/navigability win, not a load-time one.

State moves into a route-scoped `AnalyticsStore`. Routed children can't read signals off a parent component, and scoping it to the route rather than `root` preserves the old behavior of discarding data when you leave (no stale figures on a later visit). Every loader is idempotent, so each tab requests what it needs and stays self-sufficient when deep-linked, without refetching.

## One deviation worth calling out

The brief said to **hide** the Portfolio tab when portfolio tracking is unconfigured. I didn't. `HUBS` is a static registry, so hiding it would mean making the nav config depend on fetched data — and a tab bar that grows a tab after an async load is a worse experience than a stable one. The tab always renders and explains itself when tracking is off. Easy to change if you'd rather have it hidden.

## Tests

The old spec's 8 heading-based assertions were tested through the page component; they follow the code into per-tab specs, sharing the existing fixtures via `testing/analytics-fakes.ts`. The shell spec now covers what the shell actually owns — KPI composition, the pay-period toggle, and that `market` is *not* fetched eagerly.

## Verification

**518/518** tests pass, `ng lint` clean, production build succeeds, prettier clean on touched files.


StevSant (commented):
## Code review — #227

**Correct / well handled**
- Route registration is clean: `AnalyticsStore` is provided at the `analytics` route (`app.routes.ts:99-101`), one instance shared by the shell and all five lazy-loaded children, with `{ path: '', redirectTo: 'pay', pathMatch: 'full' }` for the default tab — deep links to each tab resolve correctly.
- Data-fetch split matches the PR description exactly and is verified in tests: `comp`/`funnel`/`focus` are fetched eagerly in `AnalyticsComponent.ngOnInit` (feed the KPI strip), while `market`, `skillGap`, `engagement` are deferred to their own tab's `ngOnInit` via `AnalyticsStore.loadMarket()/loadSkillGap()/loadEngagement()`. Each loader is guarded by a `*Requested` boolean so repeated/idempotent calls from tabs that also need headline data (e.g. `analytics-pay-tab.component.ts:26`, `analytics-fit-tab.component.ts:28`, `analytics-pipeline-tab.component.ts` calling `loadHeadline()` again) never duplicate requests — this is explicitly asserted in `analytics.component.spec.ts` ("fetches comp, funnel and focus eagerly... Deferred to the tab that needs it").
- `AnalyticsStore` scoped as a route `providers: []` entry (not `root`) correctly preserves the old behavior of discarding analytics data when the user navigates away, avoiding stale figures on a later visit.
- Old monolithic template/styles are genuinely deleted, not just hidden — `analytics.component.html` dropped from 166 lines of markup to a 3-line shell (header + KPI strip + `<router-outlet>`), and the corresponding scss shrank to just the page/header rules with no orphaned selectors left behind.
- Portfolio tab intentionally always renders per the hub's static tab bar (documented deviation from the brief), and both the configured/empty/unconfigured states are covered by tests (`analytics-portfolio-tab.component.spec.ts`).
- Shared fixtures (`testing/analytics-fakes.ts`) are reused across shell + all 5 tab specs, keeping the split honest instead of re-deriving mocks per file.
- `hubs.const.ts` and `hub-for-url.spec.ts` correctly updated for the 5 new tab paths under `/dashboard/analytics/*`.

- 🟠 `frontend/src/app/pages/analytics/analytics-tab-shared.scss:12-21` — `.analytics-grid` (the 2-column grid container, with its own responsive breakpoint) is dead: it's defined but no template in this PR wraps any sections in that class anymore (confirmed via grep across all analytics HTML). Previously it paired Pay+Focus and Market+Skill-gap side by side. Now Pay tab's two cards (comp benchmark, market salary range) and Fit tab's two cards (focus, skill gap) are direct siblings with no grid ancestor, so they'll stack full-width instead of rendering as two columns. `.analytics-card-wide { grid-column: 1 / -1; }` (line 123) is similarly inert without a grid ancestor — harmless for the wide cards since they already read full-width as plain blocks, but it's dead styling. Worth confirming this stacked layout is the intended visual for Pay/Fit before merge, and removing the now-dead `.analytics-grid`/`grid-column` rules either way.
- 🟡 `frontend/src/app/pages/analytics/analytics-tab-shared.scss:11-19,264` — stale comments left over from the pre-split layout: "Two equal columns (Pay | Focus, Market | Skill gap)" no longer describes anything (Pay and Focus are now separate routes, not two columns of one grid), and the file ends with a dangling comment, `/* Market context demoted into a collapsible footer. */`, referencing the old `<details class="analytics-context">` block that this PR removed — there's no code after it.
- 🟡 `frontend/src/app/pages/analytics/analytics-row-caps.ts` — bundles three unrelated constants (`PERCENT`, `MARKET_ROW_CAP`, `ENGAGEMENT_ROW_CAP`) in one file, which is a deviation from the repo's "one constant per file" convention. Minor since they're tightly related "shared numeric caps for the analytics tabs," but flagging per the stated convention.

**Verified**
- Route config in `app.routes.ts` (child routes, `redirectTo`/`pathMatch: 'full'`, lazy `loadComponent` for shell and all 5 tabs).
- Data-fetch hoisting vs. per-tab deferral, and idempotency of repeated loader calls, both in the store implementation and in the shell/tab specs.
- Full deletion of the old in-page tab bar, `<details>` "Market context" collapsible, and their markup/styles — no leftover HTML, only the two dead-CSS issues above.
- `hubs.const.ts` / `hub-for-url.ts` / `hub-for-url.spec.ts` updated consistently for the new per-tab paths.
- Accessibility: no `role="tab"` misuse introduced by this PR (the hub tab bar itself is out of scope for this diff); the in-tab period toggle keeps its existing `role="group"` + `aria-label`.
- Test coverage: old page's 8 heading-based assertions now live as focused specs per tab (`analytics-pay-tab`, `analytics-fit-tab`, `analytics-pipeline-tab`, `analytics-market-tab`, `analytics-portfolio-tab`), plus a shell spec that specifically asserts KPI composition and that `market` is *not* fetched eagerly.

Review by Claude Code

PR #228, opened 2026-08-09
feat(core): add a shared WAI-ARIA combobox for the growing dropdowns

StevSant:
Item 6 of the UI restructure. **Stacked on #227** (→ #226 → #224).

## Why these three

Exactly three pickers select from lists that grow with the data, and all three were native `<select>`s you had to scan by eye:

| Where | Problem |
|---|---|
| `job-filters` source | grows with every adapter and ATS portal ingested |
| `matching` job picker | **only fetched 25 rows**, because a longer list was unusable |
| `outreach` application picker | up to 100 entries |

The other ~30 `<select>`s stay native on purpose — short static enums (tone, language, work mode, status, sort, provider) where native wins on mobile and with assistive tech.

## The component

`core/components/combobox/`, implementing the ARIA 1.2 editable combobox with list autocomplete:

- substring filter, case-insensitive
- Arrow up/down with wraparound, Home/End, Enter to choose, Escape to cancel back to the current selection
- click-outside to dismiss
- `aria-expanded` / `aria-controls` / `aria-activedescendant` kept in sync; per-instance ids so several can coexist
- an `opened` output so hosts can lazy-load on first open — that's how matching keeps deferring its job fetch

13 specs cover the contract, keyboard model, filtering, and the disabled case.

**One lint suppression, with reasoning.** The a11y rule wants a keyboard handler on the clickable `<li role="option">`. In this pattern options are deliberately *not* focus targets — the input holds DOM focus and points at the active row via `aria-activedescendant`, and Arrow/Enter/Escape are handled there. Adding per-option key handlers would be wrong, so the rule is disabled on that line with the rationale inline. `tabindex="-1"` satisfies the focusable rule legitimately.

## Verification

**531/531** tests pass, `ng lint` clean (including the a11y step), production build succeeds, prettier clean on touched files.

### DIFF
--- a/frontend/src/app/core/components/combobox/combobox-option.model.ts
+++ b/frontend/src/app/core/components/combobox/combobox-option.model.ts
@@ -1,0 +1,5 @@
+/** One choice in a {@link ComboboxComponent}. `label` is what the user types against. */
+export interface ComboboxOption {
+  readonly value: string;
+  readonly label: string;
+}
--- a/frontend/src/app/core/components/combobox/combobox.component.html
+++ b/frontend/src/app/core/components/combobox/combobox.component.html
@@ -1,0 +1,62 @@
+<div class="combobox" [class.open]="open()">
+  <input
+    #input
+    type="text"
+    class="combobox-input"
+    role="combobox"
+    autocomplete="off"
+    [id]="id"
+    [attr.aria-expanded]="open()"
+    [attr.aria-controls]="id + '-listbox'"
+    aria-autocomplete="list"
+    [attr.aria-activedescendant]="activeOptionId()"
+    [attr.aria-label]="ariaLabel() || null"
+    [attr.aria-labelledby]="labelledBy() || null"
+    [attr.placeholder]="placeholder() || null"
+    [disabled]="disabled()"
+    [value]="displayValue()"
+    (input)="onInput($event)"
+    (focus)="onFocus()"
+    (keydown)="onKeydown($event)"
+  />
+
+  <button
+    type="button"
+    class="combobox-toggle"
+    tabindex="-1"
+    [attr.aria-label]="open() ? 'Close options' : 'Open options'"
+    [attr.aria-expanded]="open()"
+    [attr.aria-controls]="id + '-listbox'"
+    [disabled]="disabled()"
+    (click)="onToggle()"
+  >
+    <span aria-hidden="true">▾</span>
+  </button>
+
+  <ul class="combobox-listbox" role="listbox" [id]="id + '-listbox'" [hidden]="!open()">
+    @for (option of filtered(); track option.value; let i = $index) {
+      <!--
+        In the ARIA combobox pattern the input keeps DOM focus and points at the
+        active row via aria-activedescendant, so options are deliberately not
+        keyboard targets — Arrow/Enter/Escape are handled on the input above.
+        The click handler is the mouse affordance for that same behaviour.
+      -->
+      <!-- eslint-disable-next-line @angular-eslint/template/click-events-have-key-events -->
+      <li
+        class="combobox-option"
+        role="option"
+        tabindex="-1"
+        [id]="optionId(i)"
+        [class.active]="i === activeIndex()"
+        [class.selected]="option.value === value()"
+        [attr.aria-selected]="option.value === value()"
+        (mouseenter)="activeIndex.set(i)"
+        (click)="select(option)"
+      >
+        {{ option.label }}
+      </li>
+    } @empty {
+      <li class="combobox-empty" role="presentation">{{ emptyText() }}</li>
+    }
+  </ul>
+</div>
--- a/frontend/src/app/core/components/combobox/combobox.component.html
+++ b/frontend/src/app/core/components/combobox/combobox.component.html
@@ -33,13 +33,30 @@
     <span aria-hidden="true">▾</span>
   </button>
 
-  <ul class="combobox-listbox" role="listbox" [id]="id + '-listbox'" [hidden]="!open()">
+  <!--
+    Pressing the mouse inside the list must not move DOM focus off the input:
+    the option would take focus, then close() hides the listbox, and a focused
+    node that stops being rendered drops focus to <body>. Suppressing the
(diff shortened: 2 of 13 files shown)


StevSant (commented):
## Code review — #228

**Correct / well handled**
- `combobox.component.ts` implements a solid keyboard model: Down/Up open the closed list, wraparound navigation (`move()`), Home/End clamp to the filtered bounds, Enter commits and closes, Escape reverts without emitting, Tab closes without eating the tab. All verified by the specs, not just implied.
- `id`/`aria-controls`/`aria-activedescendant`/listbox id are all derived from a module-level `nextId` counter, so multiple instances on one page (job-filters + matching + outreach, three at once on a hypothetical shared page) get collision-free ids — no hardcoded id bug.
- Click-outside-to-close is done via `document:click` rather than `(blur)`, and the option `<li>` is inside the host element, so the outside-click listener never fires before the option's own `(click)` handler — that avoids the classic "blur closes list before click registers" race for the *outside-click* path.
- Call sites (`job-filters`, `matching`, `outreach`) preserve prior behavior: placeholder text, empty/`manual` default option, and the `matching` page's lazy job-fetch-on-first-open is now driven by the new `opened` output instead of `(focus)`, with the same one-shot guard semantics as before.
- `job-filters.component.spec.ts` was properly migrated to drive the new combobox through its actual DOM (focus to open, click an option), not just asserting on the underlying signal.
- The one lint suppression (`click-events-have-key-events` on the option `<li>`) is inline-commented with correct reasoning: options are intentionally not keyboard targets in this pattern, the input owns keyboard handling via `aria-activedescendant`.

- 🟠 `frontend/src/app/core/components/combobox/combobox.component.html:47-59` and `combobox.component.ts` — `aria-selected` is bound to `option.value === value()` (the last *committed* selection), not to the option currently referenced by `aria-activedescendant` (`activeIndex()`). Per the ARIA APG "Editable Combobox With List Autocomplete" pattern, `aria-selected="true"` should track whichever option is presently active/highlighted during arrow-key navigation — it's set on the option matched by `aria-activedescendant` and cleared as the user moves. As written, a screen-reader user arrowing through the list will never hear "selected" on the option they're currently on (unless it happens to equal the previously chosen value), which breaks the AT feedback loop this pattern exists to provide. Fix: drive `aria-selected` off `i === activeIndex()` (or additionally reflect it there), not off `value()`.
- 🟠 `frontend/src/app/core/components/combobox/combobox.component.html:44-59` — clicking an option can shift DOM focus away from the input before `select()` runs. The `<li role="option" tabindex="-1">` carries a `tabindex` attribute, and elements with any `tabindex` (including `-1`) are click-focusable per the HTML spec — browsers will move focus to the `<li>` on `mousedown`. `select()` then calls `close()`, which sets `open` to `false`, and the listbox gets `[hidden]`. A focused node that becomes not-rendered loses focus back to `<body>` per the HTML spec, so after a *mouse* selection the input can end up not focused at all (unlike keyboard selection, where the input never loses focus). This is the well-known "option steals focus" gotcha the reference ARIA combobox implementations guard against with `event.preventDefault()` in a `mousedown`/`pointerdown` handler on the option. None of `(mousedown)`/`(pointerdown)` here calls `preventDefault()`. Recommend adding `(mousedown)="$event.preventDefault()"` on the `<li>` (or the `<ul>`) to keep DOM focus on the input through a mouse click, matching the keyboard path's focus retention. Not caught by the existing spec because `optionEls(fixture)[0].click()` synthesizes only a `click` event, not the `mousedown` that actually triggers the browser's focus shift, so the test can't observe the regression.
- 🟡 `frontend/src/app/core/components/combobox/combobox.component.html:41` — `<li class="combobox-empty" role="presentation">{{ emptyText() }}</li>` is fully removed from the accessibility tree, so the "No matches"/`emptyText()` text is never announced to screen-reader users when a filter yields zero results; sighted users see it, AT users get silence. The official APG list-autocomplete reference example has the same gap (no live region for result count/no-results either), so this isn't a compliance regression against the pattern being followed, but it's worth a follow-up: e.g. a visually-hidden `aria-live="polite"` region announcing the match count or "No matches" would close a real UX gap for screen-reader users, since it's genuinely absent everywhere in this component.
- 🟡 `combobox.component.ts` has no `ControlValueAccessor` — it's a plain `input()`/`output()` pair rather than a form-integrated control. Not a bug today (none of the three call sites use `formControlName`/`ngModel` with it — `job-filters`/`outreach`/`matching` all bind `[value]`/`(valueChange)` directly), but flagging so a future host that needs `disabled`/`touched`/`dirty` propagation through Reactive Forms doesn't assume it's there.

**Verified**
- `git diff origin/feat/analytics-routed-tabs...origin/feat/shared-combobox` reviewed in full (13 files, ~630/58 +/-); confirmed this diff only contains this PR's changes, not #227/#226/#224's.
- Read the full new versions of `combobox.component.ts`, `.html`, `.scss`, `.spec.ts`, `index.ts`, `combobox-option.model.ts` via `git show` (not the working tree).
- Confirmed `index.ts` re-exports both `ComboboxComponent` and the `ComboboxOption` type from the package, matching the repo's "import from the contextual package" convention.
- Confirmed `matching.component.ts`'s `applyJobIdFromQuery()` still prepends a deep-linked job into the `jobs` signal directly (independent of `ensureJobsLoaded`), so `jobOptions()`/`selectedLabel` resolve correctly for a `?job_id=` deep link even before the combobox has been opened — no regression from the native-select version.
- Confirmed no other spec files (`matching.component.spec.ts`, `outreach.component.spec.ts`) query the DOM `<select>` directly — they exercise the component's signals/methods, so migrating those templates to `app-combobox` doesn't break them.
- Checked ARIA semantics against the W3C ARIA APG "Editable Combobox With List Autocomplete" reference pattern (fetched via WebFetch/WebSearch) for the `aria-selected` and result-count-announcement questions above; `role=combobox`/`aria-expanded`/`aria-controls`/`aria-autocomplete=list`/`role=listbox`/`aria-activedescendant` wiring itself matches the pattern correctly.
- Could not run `ng lint`/tests locally in this read-only review (per the git-mutation restriction); relying on the PR description's stated "531/531 pass, `ng lint` clean" plus static reading of the diff.

Review by Claude Code

PR #235, opened 2026-08-09
docs: record the frontend Node 22 and prettier check-list traps

StevSant:
Docs-only. Branched off `main`, independent of the restructure PRs (#225–#228).

Two toolchain traps cost real time during the UI restructure, and neither is discoverable from its error message:

**Node version.** Under a newer default Node, `npm test` dies with `[vitest-pool]: Failed to start forks worker` / "Timeout waiting for worker to respond". That reads like a broken test, not a runtime mismatch. `fnm exec --using=22` also doesn't work from a non-interactive shell, so the `PATH` form is documented instead.

**Prettier file lists.** Building one from `git status --porcelain` silently skips every file inside a newly created directory, because git collapses untracked directories into a single `?? path/to/dir/` entry. Local formatting passes; CI fails. This is exactly what happened on #226–#228 — all three needed a follow-up formatting commit.

Also lists `npx ng lint` and `npm run format:check` alongside the other frontend commands, since neither runs as part of `npm test` or `npm run build` but both gate CI.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -47,7 +47,29 @@ npm run build                            # production build
 npm test                                 # Vitest via ng test
 npm test -- --include "**/foo.spec.ts"   # single spec file
 npm test -- --filter "test name"         # single test by name
+npx ng lint                              # NOT run by npm test or npm run build — CI runs it
+npm run format:check                     # prettier gate, also CI-enforced
 ```
+
+**Node version.** The frontend needs Node 22 (via `fnm`). Under a newer default Node,
+`npm test` fails with `[vitest-pool]: Failed to start forks worker` / "Timeout waiting
+for worker to respond" — an opaque error that looks like a broken test rather than a
+runtime mismatch. `fnm exec --using=22` does not work from a non-interactive shell; put
+it on `PATH` instead:
+
+```bash
+export PATH="$HOME/AppData/Roaming/fnm/node-versions/v22.23.1/installation:$PATH"
+```
+
+**Prettier before pushing.** CI runs `prettier --check "src/**/*.{ts,html,css,scss,json}"`.
+Two traps when checking locally:
+
+- Never build the file list from `git status --porcelain` — it collapses untracked
+  directories into one entry, so every file inside a newly created folder is silently
+  skipped and only CI catches it. Format all of `src`, or list files from
+  `git diff --name-only origin/main...HEAD`.
+- A bare `npx prettier --check "src/**/*"` flags all ~435 files on a CRLF checkout while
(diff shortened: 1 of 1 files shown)


StevSant (commented):
## Code review — #235

**Correct / well handled**
- The Node 22 requirement matches CI (`.github/workflows/ci.yml` `setup-node@v7` `node-version: 22`).
- `npx ng lint` and `npm run format:check` claims are accurate: CI runs both as separate steps ("Lint (angular-eslint...)" and "Format check (prettier)"), and neither is invoked by `ng test` or `ng build`.
- The exact prettier command quoted (`prettier --check "src/**/*.{ts,html,css,scss,json}"`, CLAUDE.md:64) is a verbatim match of the `format:check` script in `frontend/package.json:15`.
- The CRLF trap claim checks out: `frontend/.prettierrc` sets no `endOfLine` override (defaults to `lf`), and there's no `.gitattributes`, so a Windows checkout with `core.autocrlf=true` really will produce CRLF locally while CI (Linux) sees LF — `--end-of-line auto` is the correct workaround.
- The untracked-directory collapse behavior of `git status --porcelain` is real git behavior, and it's a plausible root cause for "local passes, CI's format:check fails" — a genuinely non-obvious trap.
- Placement is consistent with the file's existing convention (the backend section already has a "Quirk:" callout for the `uv run` trampoline issue); `docs/superpowers/` is reserved for feature specs/plans, not toolchain notes, so CLAUDE.md is the right home.
- English throughout — correct for this non-Patrimore repo.
- Markdown hygiene is clean: fenced code block uses `bash`, matches surrounding style, no broken links.

- 🟡 `CLAUDE.md:61` — the `fnm` PATH snippet hardcodes a specific patch version (`v22.23.1`) and a Windows-profile-relative path. That's this machine's exact install directory; another contributor's `fnm`-managed Node 22 will live under a different version folder (e.g. `v22.20.0`), so the command won't work verbatim if copy-pasted. Worth a one-line caveat, e.g. "adjust the version segment to match `fnm list`'s output" — otherwise a reader may think `v22.23.1` is a project-pinned requirement rather than an example.

**Verified**
- Cross-checked all quoted commands/scripts (`format:check`, `ng lint`, `prettier --check` pattern) against `frontend/package.json` and `.github/workflows/ci.yml` on `origin/main`.
- Confirmed no `.nvmrc` / `engines` field exists in the repo, so this doc is the only place recording the Node 22 requirement for local dev — not a duplicate of an existing note.
- Confirmed no contradiction with the existing "Frontend (run from `frontend/`)" command block — the new content is additive, not overlapping.

**Verdict:** LGTM ✅

Review by Claude Code

PR #244, opened 2026-08-19
fix: honest sign-in errors, unwasted quick-score batches, and an observable closure sweep

StevSant:
## What & why

Three fixes that came out of chasing an "Invalid credentials" error that had nothing to do with credentials.

**1. `fix(login)` — the sign-in form told the same lie for every failure.**
`AuthService.login()` is two requests (`POST /auth/login`, then `GET /auth/me` to read the session back), and `refreshSession()` collapses any probe failure into `null`. The component rendered both that `null` and every thrown error as "Invalid credentials", so a 429 from the login rate limiter (5 attempts / 15 min), a 503 from an unconfigured backend, a client-side timeout and an unreachable API were all indistinguishable from a wrong password. Adds `mapLoginError()`, modeled on the existing `mapLlmError`: per-status messages for 401/429/503/408/0, falling back to the server's `detail` and then the status number so an unmapped failure stays identifiable. The `null` branch now states that the credential was **accepted** and only the session probe failed, since a rejected credential arrives on the error path.

**2. `fix(ingestion)` — the job-list toggles were stranded.**
`.toggle-label` carried `margin-left: auto`, written when the sort bar held one toggle beside a select. With two toggles and no select, each claimed its own auto margin, so the free space was split between them: "Show closed" landed mid-bar, "Show low-quality" at the far right, and both were squeezed until their labels wrapped mid-word ("Show low- / quality"). They now sit in a `.sort-toggles` flex group that owns the auto margin and wraps as a unit.

**3. `fix(llm)` — every full quick-scoring batch was being thrown away.**
`llm_usage_log` recorded six `match_quick_scorer` calls with `output_tokens` of exactly **2048** — pinned at `llm_default_max_tokens`. The quick scorer batches a whole page (`match_quick_batch_size`, 20) into one call, and a full page needs ~2.7k output tokens. Past the cap the response is cut mid-array, `extract_json` fails, and `_parse` discards the *entire* batch: all 20 jobs silently fall back to their heuristic score while the truncated completion is billed in full.

`llm_config_service` already excluded `match_quick_scorer` from the small classifier cap for exactly this reason, but the "larger default" it fell back to is shared with single-shot features and was never sized for a batch. Adds a third tier, `llm_batch_max_tokens` (6144), for the features whose one response carries many items — `match_quick_scorer` and `match_dimension_scorer`, where truncation drops the tail dimensions rather than shortening a verdict. A cap is a ceiling, not a charge, so this bills *less* than the status quo: the tokens already being paid for come back as usable scores. An explicit admin-set `max_tokens` still wins.

Also logs the response length and **tail** on an unparseable quick-scoring response — the old 160-char head can't distinguish a truncated response (well-formed from the front) from malformed output, which is what sent this investigation after the markdown fence when `extract_json` had handled fences all along.

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ Feature
- [ ] ♻️ Refactor
- [ ] 📝 Docs
- [ ] 🔧 Chore / tooling

## How it was tested

- `npm test` → **710 passed / 108 files**, coverage gates met. `npx ng lint` and `prettier --check` clean.
- `uv run python -m pytest` → **1773 passed, 6 skipped**. The only failures are the two documented `backend/.env` artifacts (`test_app_mode_defaults_to_local`, `test_public_symbols_importable`, both from `APP_MODE=production`); re-running from the repo root so `backend/.env` is never resolved gives 27/27 on those files plus the changed admin tests.
- `uv run ruff format --check` / `ruff check .` clean.
- New tests: 8 cases for `mapLoginError`, 3 updated/added `LoginComponent` cases, and admin-resolver cases for the batch tier (batched features get 6144, single-shot stay at 2048, all three tiers honor custom values, admin-set `max_tokens` wins).
- Verified against the running stack: `POST /api/auth/login` → 200 + `Set-Cookie`, `GET /api/auth/me` → 200 with a session and a clean 401 without one.

**Not verified visually:** the toggle-group layout. That bar only renders with `jobs().length > 0`, which needs a signed-in session, so it was reasoned from the CSS rather than screenshotted.

## Checklist

- [x] Change is focused — one logical concern. *(Three, actually: independent one-file fixes found in one debugging session. Say the word and I'll split them.)*
- [x] Tests added/updated and passing (`uv run python -m pytest`, `npm test`).
- [x] Linters pass (`uv run ruff check .`, `npx ng lint`).
- [x] Commits follow [Conventional Commits](https://www.conventionalcommits.org/) (`type(scope): …`).
- [x] Docs / README / `.env.example` updated if behavior or setup changed — `LLM_BATCH_MAX_TOKENS` documented in `.env.example`.
- [x] No hardcoded URLs, keys, or thresholds — the new cap goes through `config/groups/llm.py` + `.env`.

https://claude.ai/code/session_01WKK8FRcSB3HCuezPUNm1GS

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -127,6 +127,11 @@ LLM_DEFAULT_MAX_TOKENS=2048
 # job_quality_classifier, application_skill_extractor, preference_explanation)
 # that return a short verdict rather than long-form text.
 LLM_CLASSIFIER_MAX_TOKENS=512
+# Output cap for features whose one response carries a whole batch (the job
+# quick scorer, the 6-dimension scorer). Must cover the widest batch: a cut-off
+# response is unparseable, not merely short, so the entire batch is thrown away.
+# This is a ceiling, not a charge — only tokens actually emitted are billed.
+LLM_BATCH_MAX_TOKENS=6144
 # Cap on extracted CV/resume text (chars) sent to the LLM structured-extraction
 # prompt during PDF upload (does not truncate the stored raw text).
 CV_PARSE_CHAR_LIMIT=20000
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -165,6 +165,15 @@ JOB_CLOSURE_MISS_THRESHOLD=2
 # URL-probe revalidation sweep cadence + cost controls (feed/search sources).
 JOB_REVALIDATION_INTERVAL_HOURS=24
 JOB_REVALIDATION_BATCH=100
+# Sources the URL-probe sweep skips (comma-separated). hn_hiring/csv have no
+# per-URL closure signal; himalayas/jobicy answer every probe with a 403 or a
+# bot challenge, so probing them burns one request per job per sweep and closes
+# nothing. Sources with a declared expiry_date close via the expiry pass.
+JOB_REVALIDATION_EXCLUDED_SOURCES=hn_hiring,csv,himalayas,jobicy
+# URL substrings that mark a redirect target as an expired-listing landing page,
+# letting the probe close the job without fetching that page. A redirect to the
+# site root is detected structurally and needs no entry here.
+JOB_REVALIDATION_EXPIRED_REDIRECT_MARKERS=trk=expired_jd_redirect
 JOB_REVALIDATION_CONCURRENCY=2
 JOB_REVALIDATION_DELAY=1.0
 # SSRF hardening: probe targets derive from ingested (attacker-influenceable)
--- a/backend/src/hiresense/admin/domain/llm_config_service.py
+++ b/backend/src/hiresense/admin/domain/llm_config_service.py
@@ -13,19 +13,28 @@ logger = logging.getLogger(__name__)
 # Feature keys whose output is a short verdict — a label, a confidence score,
 # a brief extraction — rather than long-form or batched generation. These get
 # the smaller `llm_classifier_max_tokens` default instead of
-# `llm_default_max_tokens`. match_quick_scorer is deliberately NOT included
-# here even though it "classifies": it returns batched per-job JSON for up to
-# match_quick_batch_size jobs in a single call, so it needs the larger default
-# to avoid truncating the tail of the batch. match_dimension_scorer is
-# excluded for the same reason: one response carries 6 dimensions, each with
-# a score and a rationale, and truncation would silently drop the tail
-# dimensions rather than just shortening one verdict.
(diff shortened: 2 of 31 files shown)


StevSant:
Two more commits, from the "it's stuck" report on the closure sweep. It was not stuck — a full sweep probes ~2,800 open listings in throttled 100-job chunks and runs for tens of minutes. But two real defects were hiding inside that.

### `fix(ingestion)` — dead-end redirects were re-probed forever

Boards rarely 404 a removed listing. WeWorkRemotely 301s it to the site root, LinkedIn to a generic search carrying `trk=expired_jd_redirect`. The sweep followed the hop, spent a second request on the landing page, then ran the closure classifier against a **homepage** body — which matches no marker, so the verdict was UNKNOWN and the job stayed open, to be re-probed on every future sweep at two requests a go. The logs show `weworkremotely.com/` fetched hundreds of times in one sweep.

Now the redirect target itself is recognised and the job closed without fetching the landing page. Conservative by design, since a false positive closes a live listing: only an exactly-generic final path or a configured marker counts. getonbrd's constant `/jobs/<slug>` → `/jobs/<category>/<slug>` canonical rewrite is *not* a dead end and is still followed — there's a test pinning that.

Also moved the probeable-source exclusions from the composition wiring into config and added `jobicy`: it 403s every probe, exactly like the already-excluded `himalayas`, so its 217 open jobs each burned a request per sweep to produce UNKNOWN.

### `feat(ingestion)` — the sweep is now observable

The banner said "still scanning… more may drop off shortly" and nothing would ever clear or update it, because the UI polled the **job list** on a fixed 8-tick / 2-minute budget — giving up ~40 minutes early, and re-running the whole scored feed (LLM calls included) on each tick to learn nothing had changed.

`GET /ingestion/revalidate/status` answers from in-memory counters with no DB access, so it is cheap to poll. The store polls that instead, shows a real ratio, stops when the sweep reports completion, and re-reads the list only when the closed count actually moves.

### Testing

`uv run python -m pytest` → **1786 passed, 6 skipped** (+13). `npm test` → **712 passed / 108 files**. `ruff check`/`ruff format --check`, `ng lint` clean. The two `test_mode`/`test_package_layout` failures are the documented `backend/.env` artifact — 21/21 from the repo root where `.env` isn't resolved.

Not verified live: the sweep's end-to-end behavior against real boards. The redirect logic is covered by unit tests at both the helper and service level, but the first real sweep after this lands is worth watching.

PR #249, opened 2026-08-31
fix(runner): detect challenges in shadow roots and cross-origin frames

StevSant:
Closes the residual gap that the review of #248 identified and that #248's docstring documented rather than fixed.

## The gap

`page.content()` serializes neither shadow roots nor cross-origin frame contents. A captcha rendered into either was invisible to the DOM serializer, so the agent proceeded on a page a human should have taken over — **the one direction this subsystem must never fail in**. `grecaptcha.render(el, {...})` into a custom element is a normal enterprise integration, not an exotic case.

## The fix

Detection moves to where the information actually exists. `BrowserDriver` gains `challenge_present()`; `PlaywrightDriver` answers it by:

- enumerating `page.frames`, which **includes cross-origin children**, and applying the same host/path/size rules the serializer uses; and
- querying `.g-recaptcha, .h-captcha, .cf-turnstile` with a locator, which **pierces open shadow roots**, then checking visibility.

`AgentLoop` ORs the result into the observation and skips the probe when the HTML already answered, so the common path costs nothing extra.

Degrades safely: a driver written before this method existed, or one that raises, yields "no challenge" rather than aborting the run.

## Evidence

```
real Greenhouse form  -> challenge_present = False   (no regression; 5 frames incl. the enterprise anchor)

shadow-DOM widget:
  serializer sees challenge : False   <- the gap
  driver probe sees it      : True    <- closed
```

## Also in this PR

A page with **no inputs at all** no longer escalates with "Every field is filled but no submit control was found on the page". A closed posting commonly 302s to the company careers index and lands exactly there — I hit this live with job `7902104`. The old message sends the reader hunting for a bug that does not exist; it now says the page has no application form and names the URL it landed on.

## Test plan

- [x] `2083 passed` under CI conditions (+7 new tests).
- [x] New `tests/unit/runner/test_driver_challenge_probe.py`: driver-reported challenge reaches the server, clean pages stay clean, probe failure is non-fatal, a probe-less driver still works, and the probe is skipped when the HTML already flagged it.
- [x] Two new `FormAgentService` tests pin the formless-page message and confirm the filled-form-no-submit message is unchanged.
- [x] Verified live against a real Cloudflare/Greenhouse posting and a synthetic shadow-DOM widget (output above).
- [x] pyright, ruff clean.

Note: `ruff format` rewrote line endings on 30 unrelated files; those were restored so this PR touches only the 6 files it should.

Still not covered, and deliberately: a **closed** shadow root is unreachable by any client-side means, and a challenge that only appears after submit will surface as a failed submit rather than an escalation.

https://claude.ai/code/session_01EYZyM5Pm4USVavcJeqjZuA

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -60,6 +60,7 @@ testpaths = ["tests"]
 asyncio_mode = "auto"
 markers = [
     "pgvector: requires a live Postgres+pgvector DB (docker compose up db); skipped unless explicitly selected",
+    "playwright: drives a real headless browser against synthetic pages; skipped unless explicitly selected",
 ]
 
 [tool.ruff]
--- a/backend/src/hiresense/runner/agent_loop.py
+++ b/backend/src/hiresense/runner/agent_loop.py
@@ -36,6 +36,11 @@ class AgentLoop:
                 url=await self._driver.url(),
                 title=await self._driver.title(),
             )
+            # The serializer only sees serialized HTML. Ask the driver about the
+            # places it cannot reach -- shadow roots and cross-origin frames --
+            # so a challenge rendered there still stops the run.
+            if not observation["captcha_detected"] and await self._challenge_present():
+                observation["captcha_detected"] = True
             action = await self._client.observe(attempt_id, observation)
             kind = action.get("kind")
 
@@ -62,6 +67,17 @@ class AgentLoop:
             "failed",
             {"reason": f"Exceeded the {self._max_steps}-step ceiling without reaching submit"},
         )
+
+    async def _challenge_present(self) -> bool:
+        """Driver-side captcha check, tolerant of a driver that lacks it."""
+        probe = getattr(self._driver, "challenge_present", None)
+        if probe is None:
+            return False
+        try:
+            return bool(await probe())
+        except Exception:  # noqa: BLE001 - advisory signal, never fatal
+            logger.exception("runner: driver challenge probe failed")
+            return False
 
     async def _perform(self, attempt: dict, action: dict) -> None:
         kind = action.get("kind")
--- a/backend/src/hiresense/runner/agent_loop.py
+++ b/backend/src/hiresense/runner/agent_loop.py
@@ -69,15 +69,17 @@ class AgentLoop:
         )
 
     async def _challenge_present(self) -> bool:
-        """Driver-side captcha check, tolerant of a driver that lacks it."""
-        probe = getattr(self._driver, "challenge_present", None)
-        if probe is None:
-            return False
+        """Driver-side captcha check. Fails CLOSED.
+
+        A probe that cannot answer is treated as "challenge present", so the
+        attempt escalates to a human rather than proceeding to type the
+        candidate's data into a page we could not verify.
+        """
         try:
-            return bool(await probe())
-        except Exception:  # noqa: BLE001 - advisory signal, never fatal
-            logger.exception("runner: driver challenge probe failed")
-            return False
+            return bool(await self._driver.challenge_present())
+        except Exception:  # noqa: BLE001 - unverified means escalate, not proceed
+            logger.exception("runner: challenge probe failed; escalating to be safe")
+            return True
 
     async def _perform(self, attempt: dict, action: dict) -> None:
         kind = action.get("kind")
--- a/backend/src/hiresense/runner/browser_driver.py
+++ b/backend/src/hiresense/runner/browser_driver.py
@@ -27,4 +27,14 @@ class BrowserDriver(Protocol):
 
     async def text(self) -> str: ...
 
+    async def challenge_present(self) -> bool:
+        """True when a captcha challenge is on the page that HTML cannot show.
+
+        `page.content()` serializes neither shadow roots nor cross-origin frame
+        contents, so a widget rendered into either is invisible to the DOM
+        serializer. The driver sits on the other side of that boundary and can
+        see both, so it answers this question instead of the serializer.
+        """
+        ...
+
     async def close(self) -> None: ...
--- a/backend/src/hiresense/runner/challenge_probe_error.py
+++ b/backend/src/hiresense/runner/challenge_probe_error.py
@@ -1,0 +1,11 @@
+from __future__ import annotations
+
+
+class ChallengeProbeError(RuntimeError):
+    """Raised when the driver cannot determine whether a challenge is present.
+
+    Deliberately not swallowed by the driver. "I could not tell" is not the same
+    answer as "there is no challenge", and this subsystem's rule is that
+    under-escalating is the unacceptable direction -- proceeding would type the
+    candidate's data into a form that may be gated behind a captcha.
+    """
--- a/backend/src/hiresense/runner/dom_serializer.py
+++ b/backend/src/hiresense/runner/dom_serializer.py
@@ -12,30 +12,42 @@ MAX_LABEL_CHARS = 300
 
 # Hosts that serve captcha frames. Matched against an iframe's src, never
 # against the page source at large -- see _detect_captcha.
-_CAPTCHA_FRAME_HOSTS = (
+CAPTCHA_FRAME_HOSTS = (
     "recaptcha",
     "hcaptcha",
     "turnstile",
     "funcaptcha",
     "arkoselabs",
 )
+_CAPTCHA_FRAME_HOSTS = CAPTCHA_FRAME_HOSTS
 
 # The path segment of the frame that actually presents a challenge to a human
 # ("pick every bus"). reCAPTCHA/hCaptcha always embed an `anchor` frame -- the
 # badge or checkbox -- but only inject a `bframe` when a challenge is really
 # being asked. Blocking on `anchor` would block every Greenhouse posting.
-_CHALLENGE_FRAME_PATHS = ("/bframe", "/challenge")
+CHALLENGE_FRAME_PATHS = ("/bframe", "/challenge")
+_CHALLENGE_FRAME_PATHS = CHALLENGE_FRAME_PATHS
 
 # An `anchor` frame is a passive badge for invisible/v3 reCAPTCHA, but for v2 it
 # IS the "I'm not a robot" checkbox a human has to click. The rendered size is
 # what distinguishes them, and the live Greenhouse enterprise anchor carries
 # neither, so keying on these does not regress it.
-_INTERACTIVE_FRAME_SIZES = ("size=normal", "size=compact", "frame=checkbox")
+INTERACTIVE_FRAME_SIZES = ("size=normal", "size=compact", "frame=checkbox")
+_INTERACTIVE_FRAME_SIZES = INTERACTIVE_FRAME_SIZES
 
 # Container elements a captcha library renders a visible widget into. Compared
 # by EXACT class name: `g-recaptcha-response` is the hidden textarea holding the
 # token, not a widget, and must not match.
-_CAPTCHA_WIDGET_CLASSES = frozenset({"g-recaptcha", "h-captcha", "cf-turnstile"})
+CAPTCHA_WIDGET_CLASSES = frozenset({"g-recaptcha", "h-captcha", "cf-turnstile"})
+_CAPTCHA_WIDGET_CLASSES = CAPTCHA_WIDGET_CLASSES
+
+# The same rule as a CSS selector, for the driver-side probe. The
+# `data-size="invisible"` exclusion mirrors _is_invisible(): an invisible-mode
+# container is a config artifact, not a challenge, and it is what stops every
+# reCAPTCHA-protected posting from escalating. The two detectors MUST agree.
+CAPTCHA_WIDGET_SELECTOR = ", ".join(
+    f'.{name}:not([data-size="invisible" i])' for name in sorted(CAPTCHA_WIDGET_CLASSES)
+)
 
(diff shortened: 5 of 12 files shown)


StevSant (commented):
## Code review — #249

**Correct / well handled**
- The gap is real and the chosen mechanism closes it. Reproduced against headless Chromium with a synthetic page (a widget rendered into an **open** shadow root): `serialize_dom(...)["captcha_detected"] = False`, `PlaywrightDriver.challenge_present() = True`. The locator does pierce open shadow roots as claimed.
- `page.frames` is the right primitive: it is the whole frame tree (recursive, includes cross-origin children), so nested grandchild challenge frames are covered. `frame.url` is a `str` property, never `None`; the `(frame.url or "")` guard is belt-and-braces but harmless, and `about:blank` frames simply fail the host test.
- The OR-in is sound. Skipping the probe when `observation["captcha_detected"]` is already `True` cannot change the result (`x or y` with `x=True`), and it is pinned by `test_probe_is_skipped_when_the_html_already_shows_a_challenge` (`driver.probes == 0`).
- Mutating the dict before `client.observe` is safe end-to-end: `ObserveRequest.observation: PageObservation` declares `captcha_detected: bool`, so the flag survives server-side validation and reaches `FormAgentService.next_action`'s first branch (`form_agent_service.py:88`), which escalates before any answering. Verified the whole path, not just the runner half.
- `locator.nth(i).is_visible()` does **not** throw on a detached element — verified: after removing the node between `count()` and `is_visible()`, it returns `False`. So the count/visibility split is not a crash risk, only a staleness one.
- Layering is respected: `runner/` still imports nothing from any module's `domain/`/`infrastructure/`/`api/`; `playwright_driver.py` adds only `logging`.
- The `_finish` restructure is behaviourally correct. `not any(f.field_type not in _SUBMIT_TYPES for f in observation.fields)` = "no non-submit field exists", vacuously true for `fields == []`, and the early `return SubmitAction` makes the two escalation branches genuinely disjoint. The old message is preserved for the real "filled form, no submit" case and pinned by a new test.

- 🟠 `backend/src/hiresense/runner/playwright_driver.py:105-107` — **the widget probe drops the `data-size="invisible"` exclusion that #248 added, and I reproduced the false positive.**
  ```python
  locator = self._page.locator(_WIDGET_SELECTOR)   # ".g-recaptcha, .h-captcha, .cf-turnstile"
  for index in range(min(await locator.count(), 5)):
      if await locator.nth(index).is_visible():
          return True
  ```
  `dom_serializer._detect_captcha` requires `... and not _is_invisible(element)` precisely because an invisible-mode container is a config artifact, not a challenge — that guard is what stopped every Greenhouse posting escalating. The probe replaces it with Playwright's `is_visible()`, which is a *different question*. Measured:

  | page | serializer | probe |
  |---|---|---|
  | `<div class="g-recaptcha" data-size="invisible" style="256x60">badge</div>` | `False` | **`True`** |
  | `<div class="g-recaptcha" data-size="invisible"></div>` (empty) | `False` | `False` |

  Automatic invisible reCAPTCHA renders its `.grecaptcha-badge` (256×60) **inside** the `.g-recaptcha` container, which gives the container a non-empty box → `is_visible() == True`. That is row 1, i.e. the exact `GOOGLE_RECAPTCHA_INVISIBLE_KEY` shape #248 was written to tolerate. The one live Greenhouse posting you checked came back `False`, but that is a single sample of a widget whose badge placement varies with `badge:` / explicit-render config. Since the probe is OR'd in, a hit here re-escalates every posting on that ATS. Mirror the serializer: exclude `[data-size="invisible" i]` in `_WIDGET_SELECTOR`, or read `data-size` per element before accepting the hit.

- 🟠 `backend/src/hiresense/runner/playwright_driver.py:106` — **`min(count, 5)` is a silent correctness hole in the dangerous direction, and it is not hypothetical.** Reproduced: a page with six `display:none` `.g-recaptcha` stubs followed by one visible widget gives `serializer=True, probe=False`. Multiple inert widget containers are ordinary in multi-step / SPA application forms (one per step, all but the active one hidden), so "index > 5" is not exotic — and this cap is exactly what makes the shadow-DOM variant of that page slip through, which is the case this PR exists to catch. There is no perf argument for the cap: replace the whole loop with one round trip that has no cap and no staleness window between `count()` and `is_visible()`:
  ```python
  if await self._page.locator(f"{_WIDGET_SELECTOR} >> visible=true").count():
      return True
  ```
  If a cap is genuinely wanted it needs a named constant and a comment justifying the number — a bare `5` in a safety check reads as an oversight.

- 🟠 `backend/src/hiresense/runner/playwright_driver.py:109-110` + `agent_loop.py:53-55` — **a safety check that fails open, twice.** Both layers catch bare `Exception` and return `False`, so "I could not determine whether a human is being challenged" is reported to the server as "no challenge", and the agent proceeds to type the candidate's PII into the employer's form. Argued both ways: the fail-open case is that a dead page/CDP will make `driver.html()` raise *before* the probe, so probe-only failures are narrow, and aborting on an advisory signal costs availability. I do not find that persuasive here. The failures that reach this handler are precisely the ambiguous ones — a locator timeout on a busy page, a frame detaching mid-enumeration — and this subsystem's stated rule is that under-escalation is the unacceptable direction. **Escalate.** Concretely: return a tri-state (or raise a `ChallengeProbeError`) and have `AgentLoop` set `captcha_detected = True`, or emit a distinct escalation reason ("could not verify the page is challenge-free"). Whatever you choose, the current double-swallow also makes `AgentLoop`'s `logger.exception` branch unreachable for the only driver that exists — the inner handler eats everything first, so the outer one logs nothing and is dead defensive code.

- 🟠 `backend/tests/unit/runner/test_driver_challenge_probe.py` — **the new tests do not test the new detection logic.** All five drive `AgentLoop` against a fake driver whose `challenge_present` returns a canned bool; `grep -rn "PlaywrightDriver" backend/tests` returns nothing. So the file pins the *plumbing* (which is good, and which would fail against `main`) while the probe itself — the frame host/path/size rules, `_WIDGET_SELECTOR`, the visibility check, the cap — has zero coverage. Mutations the suite does not catch: `min(count, 5)` → `min(count, 0)`; `_WIDGET_SELECTOR` → `""`; inverting `if not any(host in src ...)` to `if any(...)`; deleting the frame loop entirely. Since `page.set_content()` plus a `PlaywrightDriver` with `_page` injected runs fine headless (that is how the table above was produced), a handful of opt-in `@pytest.mark.playwright` tests in the shape of the existing `pgvector` marker would pin this cheaply.

- 🟡 `backend/src/hiresense/runner/agent_loop.py:50-52` — `getattr(self._driver, "challenge_present", None)` is duck-typing around a method the Protocol now *declares*. Production has exactly one driver (`cli.py:32`, `PlaywrightDriver`) and it implements it; pyright's `include` is `src/hiresense` only, so nothing in `src` needs the fallback. What actually needs it is `tests/unit/runner/test_agent_loop.py`'s `_Driver`, which has no `challenge_present`. That is production safety code branching to accommodate a test double, and the branch it takes is "silently lose the protection". `_LegacyDriver` setting `challenge_present = None` as a class attribute is a contrivance built to enter that branch — a real pre-#249 driver would simply not define the attribute (which `getattr` also handles, so the class attribute buys nothing), and the `del self.__dict__["probes"]` / `self.probes = 0` dance is inert. Suggest: add the method to the fake in `test_agent_loop.py`, call `await self._driver.challenge_present()` directly, and delete `_LegacyDriver`. A future driver that forgets the method should be a type error, not a silently unprotected run.

- 🟡 `backend/src/hiresense/runner/playwright_driver.py:6-16` — `_CHALLENGE_FRAME_PATHS`, `_INTERACTIVE_FRAME_SIZES` and `_CAPTCHA_FRAME_HOSTS` are copied verbatim from `runner/dom_serializer.py:15-33`. Both files are in the same package, so the HTTP-only boundary is no excuse for not importing them. Adding a host to one list now silently fails to protect the other path — and the first 🟠 above is that drift already happening for the widget rule. Import the constants from `dom_serializer` (or lift them into a shared `runner/` module) so the two detectors cannot diverge.

- 🟡 `backend/src/hiresense/runner/playwright_driver.py:96` — the frame loop includes `page.main_frame`, so the substring rules are applied to the employer's own page URL rather than only to embedded frames (the serializer applies them to `<iframe src>` only). A URL like `.../apply?next=/recaptcha/api2/bframe` would match. It over-escalates rather than under-escalates so it is not dangerous, but `if frame is self._page.main_frame: continue` matches the serializer's intent.

- 🟡 `backend/src/hiresense/submission/domain/form_agent_service.py:186-194` — the new message asserts a cause ("the posting may have closed or redirected") from the absence of fields, but `observation.fields` can also be empty because the serializer *dropped* every control — `_css_path` returns `None` past `MAX_CSS_PATH_DEPTH` and such elements are discarded, which is the same failure mode the depth-60 comment describes. In that case the reader is told the posting closed when the real answer is "our selector builder gave up". Hedging the wording ("no form fields were detected on this page — the posting may have closed or redirected") keeps the improvement without asserting a diagnosis. Related: a submit-only page (`fields=[button]`, no submit-hint label) also takes this branch and is arguably a filled-form failure rather than a formless page; neither the new test (`_obs([])`) nor any other pins that shape.

- 🟡 `backend/src/hiresense/runner/playwright_driver.py:3-5` — stray blank line splitting the stdlib import block (`import logging` / blank / `from typing import Any`). Cosmetic; ruff's isort rules are not enabled so nothing catches it.

**Verified**
- `uv run python -m pytest -q` → `2081 passed, 6 skipped, 2 failed`. Both failures are the known local-`.env` pollution (`test_app_mode_defaults_to_local`, `test_public_symbols_importable`), not this PR. 2081 + 2 = **2083, matching the PR body**; +7 new tests confirmed (5 runner + 2 submission).
- `uv run ruff check` + `ruff format --check` on all six touched files: clean, as claimed.
- Headless Chromium experiments against synthetic pages via `page.set_content` (no third-party sites), injecting a real `Page` into the branch's `PlaywrightDriver`: open shadow root → probe `True` (gap closed); closed shadow root → `False` (correctly documented as out of reach); clean form → `False` (no regression); plus the two divergence rows and the detached-element result quoted above.
- Server-side acceptance of the mutated flag traced through `runner/client.py:41` → `submission/api/routes.py:74` → `ObserveRequest` → `PageObservation.captcha_detected` → `FormAgentService.next_action:88`.
- `runner/` import boundary and one-definition-per-file conventions per `CLAUDE.md` / `backend/ARCHITECTURE.md`: respected.
- PR body claims spot-checked: the frame rules *are* the serializer's (by duplication, not import); "pierces open shadow roots, then checks visibility" is accurate but omits that it drops the `data-size` guard; "degrades safely" is accurate as written but is the fail-open behaviour argued against above.

The direction is right and the plumbing is solid, but three of the four 🟠 items sit in the same place: the widget branch of the probe is looser than the serializer in one direction (invisible containers) and tighter in another (the index cap), and nothing tests it. I would not merge the `min(count, 5)` cap or the missing `data-size` guard as-is.

Review by Claude Code

StevSant:
Addressed all four 🟠 in `a5e7352d`. The two reproduced regressions were both real and both mine.

**🟠 Dropped `data-size="invisible"` guard — fixed, and the root cause was worse than one missing check.** You were right that the probe and the serializer had already drifted. Rather than re-add the guard in a second place, the rule now lives in `dom_serializer` as `CAPTCHA_WIDGET_SELECTOR` (with `:not([data-size="invisible" i])` baked in) and the driver **imports** it — along with the frame host/path/size constants, which were also copies. The two detectors can no longer disagree. Pinned by `test_invisible_widget_is_not_a_challenge`, using your 256×60 badge shape.

**🟠 `min(count, 5)` — fixed, replaced exactly as you suggested.** One uncapped `>> visible=true` query, which also removes the staleness window between `count()` and `is_visible()`. Your six-hidden-stubs case is now a test (`test_widget_beyond_the_first_five_still_blocks`), and it fails against the previous commit.

**🟠 Fails open — fixed; I agree with your reasoning.** The failures that reach that handler are precisely the ambiguous ones, and "I could not tell" is not "no challenge". `PlaywrightDriver` now raises `ChallengeProbeError`, and `AgentLoop` escalates on it. That also removes the dead outer handler you spotted — there is exactly one `except` now, and it fails closed.

**🟠 Probe untested — fixed.** New `tests/unit/runner/test_playwright_probe.py`: 12 tests driving real headless Chromium, opt-in behind a `playwright` marker gated the same way as `pgvector`, so the default run stays browser-free (`49 passed, 12 skipped`). It covers each mutation you named — the cap, the widget selector, the frame-loop inversion, and the invisible/hidden/token-textarea cases.

One thing your list did not have, which the new tests surfaced: `test_challenge_frame_blocks` initially failed because it fetched the real Google `bframe` endpoint, which returned `chrome-error://chromewebdata/` instead of its URL. Network-dependent by construction. Frames are now stubbed with `page.route`, so nothing leaves the machine.

**🟡 all taken:** main frame skipped when matching URLs; message hedged to "No form fields were detected on this page" (your point that `_css_path` may have dropped every control is right — asserting a cause was wrong); `getattr` duck-typing removed and the fake in `test_agent_loop.py` given the method, with `_LegacyDriver` deleted; constants imported rather than copied; stray blank line gone. Added the submit-only-page case you noted was unpinned.

`2084 passed, 18 skipped` on the default run; `12 passed` under `-m playwright`. pyright and ruff clean.

Thanks — the `data-size` regression in particular was a straight reintroduction of the bug #248 existed to fix, and I would not have caught it from the single live sample I had.

